PR #1, opened 2022-09-30
Implement Furi message queue

twitchyliquid64:
Heya, love youre work! I have a bunch of plugin ideas I want to work on, so if its okay with you I'd like to implement them and send them to you as PRs.

Just to get my feet wet I've implemented the low-level message queue API, I have ideas around how to wrap it up neatly in a safe generic type but I'll save that for another PR.

Tested like this:

```rust
#[repr(C)]
#[derive(Debug)]
struct msg(u8);

let queue = message_queue::alloc(8, core::mem::size_of::<msg>());

write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", message_queue::capacity(queue), message_queue::count(queue), message_queue::space(queue)).unwrap();
stdout.flush().unwrap();

message_queue::put(queue, &mut msg(1) as *mut _ as *const c_void, 9999999);

write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", message_queue::capacity(queue), message_queue::count(queue), message_queue::space(queue)).unwrap();
stdout.flush().unwrap();

let mut output = msg(0);
message_queue::get(queue, &mut output as *mut _ as *mut c_void, 9999999);

write!(&mut stdout, "Read back: {:?}\r\n", output).unwrap();
write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", message_queue::capacity(queue), message_queue::count(queue), message_queue::space(queue)).unwrap();
stdout.flush().unwrap();
```

getting output:

image

### DIFF
--- a/crates/flipperzero/src/furi.rs
+++ b/crates/flipperzero/src/furi.rs
@@ -1,5 +1,7 @@
 //! High-level bindings to Furi kernel
 
+use core::ffi::c_void;
+
 use flipperzero_sys as sys;
 pub struct Stdout;
 
@@ -35,6 +37,107 @@ pub fn sleep(duration: core::time::Duration) {
             sys::furi::delay_us(duration.as_micros() as u32);
         } else {
             sys::furi::delay_ms(duration.as_millis() as u32);
+        }
+    }
+}
+
+/// Error codes returned from Furi function calls
+#[derive(Debug)]
+pub enum Error {
+    /// General runtime error, corresponds to FuriStatusError
+    Generic,
+    /// Timeout was exceeded, corresponds to FuriStatusErrorTimeout
+    Timeout,
+    // Resource not available, corresponds to FuriStatusErrorResource
+    Resource,
+    // Parameter error, corresponds to FuriStatusErrorParameter
+    Parameter,
+    // System is out of memory, corresponds to FuriStatusErrorNoMemory
+    NoMemory,
+    // Not allowed in ISR context, corresponds to FuriStatusErrorISR
+    ISR,
+    // Unrecognized error code.
+    Other(u32),
+}
+
+impl From<u32> for Error {
+    fn from(code: u32) -> Self {
+        match i32::from_le_bytes(code.to_le_bytes()) {
+            -1 => Error::Generic,
+            -2 => Error::Timeout,
+            -3 => Error::Resource,
+            -4 => Error::Parameter,
+            -5 => Error::NoMemory,
+            -6 => Error::ISR,
+
+            _ => Error::Other(code)
+        }
+    }
+}
+
+/// Furi result type
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// MessageQueue provides a safe wrapper around the furi message queue primitive.
+pub struct MessageQueue<M: Sized> {
+    hnd: *const sys::message_queue::FuriMessageQueue,
+    _marker: core::marker::PhantomData<M>,
+}
+
+impl<M: Sized> MessageQueue<M> {
+    /// Constructs a message queue with the given capacity.
+    pub fn new(capacity: u32) -> Self {
+        Self {
+            hnd: unsafe { sys::message_queue::alloc(capacity, core::mem::size_of::<M>()) },
+            _marker: core::marker::PhantomData::<M>,
+        }
+    }
+
+    // Attempts to add the message to the end of the queue, waiting up to timeout ticks.
+    pub fn put(&self, mut msg: M, timeout_ticks: u32) -> Result<()> {
+        let code = unsafe {
+            sys::message_queue::put(self.hnd, &mut msg as *mut _ as *const c_void, timeout_ticks)
+        };
+
+        match code {
+            0 => Ok(()),
+            _ => Err(code.into()),
+        }
+    }
+
+    // Attempts to read a message from the front of the queue within timeout ticks.
+    pub fn get(&self, timeout_ticks: u32) -> Result<M> {
+        let mut out = core::mem::MaybeUninit::<M>::uninit();
+        let code = unsafe {
+            sys::message_queue::get(self.hnd, out.as_mut_ptr() as *mut c_void, timeout_ticks)
+        };
+
+        match code {
+            0 => Ok(unsafe { out.assume_init() }),
+            _ => Err(code.into()),
+        }
+    }
+
+    /// Returns the capacity of the queue.
+    pub fn capacity(&self) -> u32 {
+        unsafe { sys::message_queue::capacity(self.hnd) }
+    }
+
+    /// Returns the number of elements in the queue.
+    pub fn len(&self) -> u32 {
+        unsafe { sys::message_queue::count(self.hnd) }
+    }
+
+    /// Returns the number of free slots in the queue.
+    pub fn space(&self) -> u32 {
+        unsafe { sys::message_queue::space(self.hnd) }
+    }
+}
+
+impl<M: Sized> Drop for MessageQueue<M> {
+    fn drop(&mut self) {
+        unsafe {
+            sys::message_queue::free(self.hnd)
         }
     }
 }
--- a/crates/sys/src/message_queue.rs
+++ b/crates/sys/src/message_queue.rs
@@ -1,0 +1,30 @@
+//! Low-level bindings to Furi message queue API
+
+use core::ffi::c_void;
+
+use crate::opaque;
+
+opaque!(FuriMessageQueue);
+
+extern "C" {
+    #[link_name = "furi_message_queue_alloc"]
+    pub fn alloc(count: u32, size: usize) -> *const FuriMessageQueue;
+
+    #[link_name = "furi_message_queue_free"]
+    pub fn free(queue: *const FuriMessageQueue);
+
+    #[link_name = "furi_message_queue_put"]
+    pub fn put(queue: *const FuriMessageQueue, payload: *const c_void, timeout: u32) -> u32;
+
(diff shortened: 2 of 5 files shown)


dcoles:
Hi @twitchyliquid64!

Thanks for the PR! They look great!

At the moment I've been cramming all the `furi_` symbols under `furi.rs`, but I think long-term it's going to be better to put things under sub-modules (e.g. `furi::thread::get_name` instead of `furi::thread_get_name`). But that's easily something to do in the future.

twitchyliquid64:
Great!

Just as you posted that I finished up the `MessageQueue` type - is that okay to include in this PR? I like it a lot better than driving the `sys` crate directly.

```rust

write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();
stdout.flush().unwrap();

queue.put(msg(1), 9999999);

write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();
stdout.flush().unwrap();

let mut output = msg(0);
queue.get(&mut output, 9999999);

write!(&mut stdout, "Read back: {:?}\r\n", output).unwrap();
write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();
stdout.flush().unwrap();
```

twitchyliquid64:
> At the moment I've been cramming all the furi_ symbols under furi.rs, but I think long-term it's going to be better to put things under sub-modules

Happy to break out the message queue stuff into other sub-crates if that helps! just lemme know :)

crates/flipperzero/src/furi.rs
```diff
@@ -38,3 +40,60 @@ pub fn sleep(duration: core::time::Duration) {
...
+        Self {
+            hnd: unsafe { sys::message_queue::alloc(capacity, core::mem::size_of::<M>()) },
+            _marker: core::marker::PhantomData::<M>,
+        }
+    }
+
+    // Attempts to add the message to the end of the queue, waiting up to timeout ticks.
+    // FuriStatusOk (0) is returned if the insert is successful, otherwise FuriStatusErrorTimeout
+    // (-2) is returned.
+    pub fn put(&self, mut msg: M, timeout: u32) -> u32 {
```
dcoles, line 62: Can you use [`Duration`](https://doc.rust-lang.org/core/time/struct.Duration.html) as the type for `timeout`, rather than `u32`.

dcoles, line 62: For the high-level APIs, we should return some sort of `Result<T, Error>` rather than the error code directly.

Something probably a bit like [`nix::Error`](https://docs.rs/nix/latest/nix/type.Error.html).

crates/sys/src/message_queue.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub fn free(queue: *const FuriMessageQueue);
+
+    #[link_name = "furi_message_queue_put"]
+    pub fn put(queue: *const FuriMessageQueue, payload: *const c_void, timeout: u32) -> u32;
+
+    #[link_name = "furi_message_queue_get"]
+    pub fn get(queue: *const FuriMessageQueue, payload: *mut c_void, timeout: u32) -> u32;
+
+    #[link_name = "furi_message_queue_get_capacity"]
+    pub fn capacity(queue: *const FuriMessageQueue) -> u32;
```
dcoles, line 23: Normally Rust uses `usize` for count/size values.

In this case we're lucky, as `usize` on `thumbv7em-none-eabihf` is the same as a `u32` (see [`p:32:32`](https://github.com/rust-lang/rust/blob/6af09d2505f38e4f1df291df56d497fb2ad935ed/compiler/rustc_target/src/spec/thumbv7em_none_eabihf.rs#L17)).

crates/flipperzero/src/furi.rs
```diff
@@ -38,3 +40,60 @@ pub fn sleep(duration: core::time::Duration) {
...
+        unsafe {
+            sys::message_queue::put(self.hnd, &mut msg as *mut _ as *const c_void, timeout)
+        }
+    }
+
+    // Attempts to read a message from the front of the queue within timeout ticks.
+    // FuriStatusOk (0) is returned if the read is successful, otherwise FuriStatusErrorTimeout
+    // (-2) is returned.
+    pub fn get(&self, msg: &mut M, timeout: u32) -> u32 {
```
dcoles, line 71: Similarly, this would return something like `Result<M, furi::Error>`, rather than using a `&mut M` parameter.

crates/flipperzero/src/furi.rs
twitchyliquid64: `timeout` is actually an integer of ticks rather than an actual time. I'll change the name to `timeout_ticks` to avoid confusion.

dcoles:
@twitchyliquid64 The low-level `sys` stuff looks great, though I think we're going to need a `furi::Error` type for the high-level APIs.

If you're OK with it, I'm happy to merge the first commit and we can keep working on the high level API. Or we can keep them as one PR. Let me know what works best.

crates/sys/src/message_queue.rs
twitchyliquid64: Happy to change it. I based this off of the header file [here](https://github.com/flipperdevices/flipperzero-firmware/blob/ec57dd310a6ad9657c08d143bd82a66da532dc78/furi/core/message_queue.h), which used `uint32_t` for the value types and `FuriStatus` as the return codes. `FuriStatus` valus are all defined as `0xFFFFFFxx`, so I inferred `u32`.

crates/flipperzero/src/furi.rs
twitchyliquid64: Will do!

twitchyliquid64:
I'll give the error stuff a crack now, otherwise yeah we can merge the first commit and soldier on with the rest.

crates/flipperzero/src/furi.rs
dcoles: Oh interesting! Sorry for not seeing that in the comment.

For the high-level API, I think we would want to pass it through the equivalent of [`pdMS_TO_TICKS`](https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/4c4089b1544b590ed3b72491a15365ec020c921f/include/projdefs.h#L42), so the user doesn't have to do this calculation themselves.

crates/flipperzero/src/furi.rs
```diff
@@ -38,3 +40,108 @@ pub fn sleep(duration: core::time::Duration) {
...
+    NoMemory,
+    // Not allowed in ISR context, corresponds to FuriStatusErrorISR
+    ISR,
+    // Unrecognized error code.
+    Other(u32),
+}
+
+impl From<u32> for Error {
+    fn from(code: u32) -> Self {
+        match i32::from_le_bytes(code.to_le_bytes()) {
```
twitchyliquid64, line 65: Not sure about this. Surely these a better way to reinterpret a u32 as a i32? (preferably without a transmute or other unsafety?)

twitchyliquid64:
PTAL!

```rust

queue.put(msg(1), 9999999).ok();
write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();

let output = queue.get(9999999);
write!(&mut stdout, "Read back: {:?}\r\n", output.unwrap()).unwrap();

write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();
stdout.flush().unwrap();
```

crates/flipperzero/src/furi.rs
twitchyliquid64: Hmmm, do you know what the API would be for this in Furi?

If it turns out that the fz does frequency scaling to help manage its power budget, im not sure theres a good way to map to a duration without doing an API lookup for the frequency in the hot path.

dcoles:
I've made some adjustments to the package layout on the [`0.2-dev` branch](https://github.com/dcoles/flipperzero-rs/tree/0.2-dev). This branch also includes a [`Status` type](https://github.com/dcoles/flipperzero-rs/blob/0.2-dev/crates/sys/src/furi/base.rs#L10) that can be used as the return type for the C APIs. This is also re-exported in the main package as [`flipperzero::furi::Error`](https://github.com/dcoles/flipperzero-rs/blob/0.2-dev/crates/flipperzero/src/furi/mod.rs#L11).

twitchyliquid64:
Unfortunately `0.2-dev` is completely broken for me: the flipper crashes as soon as I load the plugin.

twitchyliquid64:
In any case heres a rebase of the message queue stuff onto `0.2-dev`: https://github.com/twitchyliquid64/flipperzero-rs/tree/tom/0.2-msgqueue (https://github.com/twitchyliquid64/flipperzero-rs/commit/d0840c3941a26f73be0878dfcad905f6d5ba60a6)

twitchyliquid64:
Okay the issue is unwrapping the return from `stdout.flush()`. Everything else works.

twitchyliquid64:
Oh, the logic is inverted:

```rust
            if sys::furi::thread::stdout_flush().is_ok() { // missing a !
                return Err(core::fmt::Error);
            }
```

twitchyliquid64:
Superceeded by https://github.com/dcoles/flipperzero-rs/pull/2

PR #2, opened 2022-09-30
0.2-dev: Implement message_queue, fix Stdout.flush

twitchyliquid64:
Rebased from #1.

```rust
write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();


queue.put(msg(69), 9999999).unwrap();
write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();

let output = queue.get(9999999);
write!(&mut stdout, "Read back: {:?}\r\n", output.unwrap()).unwrap();
write!(&mut stdout, "Queue capacity/count/space: {:?}/{:?}/{:?}\r\n", queue.capacity(), queue.len(), queue.space()).unwrap();
stdout.flush().unwrap();
```

### DIFF
--- a/crates/flipperzero/src/furi/io.rs
+++ b/crates/flipperzero/src/furi/io.rs
@@ -18,7 +18,7 @@ impl core::fmt::Write for Stdout {
 impl Stdout {
     pub fn flush(&mut self) -> core::fmt::Result {
         unsafe {
-            if !sys::furi::thread::stdout_flush().is_ok() {
+            if sys::furi::thread::stdout_flush().is_err() {
                 return Err(core::fmt::Error);
             }
         }
--- a/crates/flipperzero/src/furi/message_queue.rs
+++ b/crates/flipperzero/src/furi/message_queue.rs
@@ -8,9 +8,11 @@ pub struct MessageQueue<M: Sized> {
     _marker: core::marker::PhantomData<M>,
 }
 
+pub const TIMEOUT_NEVER: u32 = 0xFFFFFFFF; // corresponds to FuriWaitForever
+
 impl<M: Sized> MessageQueue<M> {
     /// Constructs a message queue with the given capacity.
-    pub fn new(capacity: u32) -> Self {
+    pub fn new(capacity: usize) -> Self {
         Self {
             hnd: unsafe { message_queue::alloc(capacity, core::mem::size_of::<M>()) },
             _marker: core::marker::PhantomData::<M>,
@@ -18,14 +20,17 @@ impl<M: Sized> MessageQueue<M> {
     }
 
     // Attempts to add the message to the end of the queue, waiting up to timeout ticks.
-    pub fn put(&self, mut msg: M, timeout_ticks: u32) -> Result<()> {
+    pub fn put(&self, msg: M, timeout_ticks: u32) -> Result<()> {
+        let mut msg = core::mem::ManuallyDrop::new(msg);
+
         let status = unsafe {
             message_queue::put(self.hnd, &mut msg as *mut _ as *const c_void, timeout_ticks)
         };
 
-        match status.is_ok() {
-            true => Ok(()),
-            _ => Err(status),
+        if status.is_ok() {
+            Ok(())
+        } else {
+            Err(status)
         }
     }
 
@@ -36,9 +41,10 @@ impl<M: Sized> MessageQueue<M> {
             message_queue::get(self.hnd, out.as_mut_ptr() as *mut c_void, timeout_ticks)
         };
 
-        match status.is_ok() {
-            true => Ok(unsafe { out.assume_init() }),
-            _ => Err(status),
+        if status.is_ok() {
+            Ok(unsafe { out.assume_init() })
+        } else {
+            Err(status)
         }
     }
 
@@ -60,6 +66,15 @@ impl<M: Sized> MessageQueue<M> {
 
 impl<M: Sized> Drop for MessageQueue<M> {
     fn drop(&mut self) {
+        // Drain any elements from the message queue, so any
+        // drop handlers on the message element get called.
+        while self.len() > 0 {
+            match self.get(TIMEOUT_NEVER) {
+                Ok(msg) => drop(msg),
+                Err(_) => break, // we tried
+            }
+        }
+
         unsafe {
             message_queue::free(self.hnd)
         }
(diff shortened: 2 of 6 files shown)


crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// MessageQueue provides a safe wrapper around the furi message queue primitive.
+pub struct MessageQueue<M: Sized> {
+    hnd: *const message_queue::FuriMessageQueue,
+    _marker: core::marker::PhantomData<M>,
+}
+
+impl<M: Sized> MessageQueue<M> {
+    /// Constructs a message queue with the given capacity.
+    pub fn new(capacity: u32) -> Self {
```
dcoles, line 13: We should use `usize` for "sizes", since that's the convention in the standard library so avoids a bit of casting. I was thinking of actually doing that on the C definitions, since on ARM32 `usize` = `u32`.

crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -0,0 +1,67 @@
...
+        Self {
+            hnd: unsafe { message_queue::alloc(capacity, core::mem::size_of::<M>()) },
+            _marker: core::marker::PhantomData::<M>,
+        }
+    }
+
+    // Attempts to add the message to the end of the queue, waiting up to timeout ticks.
+    pub fn put(&self, mut msg: M, timeout_ticks: u32) -> Result<()> {
+        let status = unsafe {
+            message_queue::put(self.hnd, &mut msg as *mut _ as *const c_void, timeout_ticks)
```
dcoles, line 23: Since we're handling `msg` over to C, we need to make sure that we don't drop the type at the end of this function. The easiest way to do so is using [`ManuallyDrop`](https://doc.rust-lang.org/core/mem/struct.ManuallyDrop.html):

```rust
let msg = ManuallyDrop::new(msg);
let p = &msg as &T as *const T;

// T::drop() is not called
```

crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -0,0 +1,67 @@
...
+        };
+
+        match status.is_ok() {
```
dcoles, line 26: Using an `if` block would be good here:

```rust
if status.is_ok() { Ok(()) } else { Err(status) }
```

crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Returns the number of free slots in the queue.
+    pub fn space(&self) -> u32 {
+        unsafe { message_queue::space(self.hnd) }
+    }
+}
+
+impl<M: Sized> Drop for MessageQueue<M> {
+    fn drop(&mut self) {
```
dcoles, line 62: It would be good if we also can drain the message queue and drop all the values, so we don't leak memory if the queue isn't already empty.

dcoles (changes requested):

dcoles:
Hi @twitchyliquid64. Looks like you've been busy today. 😁 

Looking pretty good! My comments are mostly around some corner cases with allocating types or custom `Drop` implementations.

twitchyliquid64:
Wow you were up late! :P

I'll address your comments and update the PR this morning. Btw I figured out how to drive the vibration motor and the status LEDs so I'll have a PR for that soon as well.

crates/flipperzero/src/furi/message_queue.rs
twitchyliquid64: Done. I use `drop` explicitly just to be verbose about whats happening.

crates/flipperzero/src/furi/message_queue.rs
twitchyliquid64: Done.

crates/flipperzero/src/furi/message_queue.rs
twitchyliquid64: TIL About `ManuallyDrop` :). Done.

crates/flipperzero/src/furi/message_queue.rs
twitchyliquid64: Done.

crates/flipperzero/src/furi/io.rs
```diff
@@ -18,7 +18,7 @@ impl core::fmt::Write for Stdout {
 impl Stdout {
     pub fn flush(&mut self) -> core::fmt::Result {
         unsafe {
-            if sys::furi::thread::stdout_flush().is_ok() {
+            if sys::furi::thread::stdout_flush().is_err() {
```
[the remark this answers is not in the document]
twitchyliquid64: Done.

crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Constructs a message queue with the given capacity.
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            hnd: unsafe { message_queue::alloc(capacity, core::mem::size_of::<M>()) },
+            _marker: core::marker::PhantomData::<M>,
+        }
+    }
+
+    // Attempts to add the message to the end of the queue, waiting up to timeout ticks.
+    pub fn put(&self, msg: M, timeout_ticks: u32) -> Result<()> {
```
dcoles, line 23: I just added [`duration_to_ticks`](https://github.com/dcoles/flipperzero-rs/blob/aa7690b733b02a47ee928ed7b74c46db75a045bc/crates/sys/src/furi/kernel.rs#L7) helper that calls `furi_ms_to_ticks` internally (on the current Flipper Zero [this is just a passthrough](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@0f9ea92/-/blob/furi/core/kernel.c?L151) 😁).

dcoles (approved):

dcoles:
> Wow you were up late! :P

Yeah, unfortunately it seems I'm most productive between the hours of 9 pm and 3 am... 🥱 

> I'll address your comments and update the PR this morning. Btw I figured out how to drive the vibration motor and the status LEDs so I'll have a PR for that soon as well.

Cool! I'm looking forward to it!

PR #3, opened 2022-09-30
0.2-dev: Implement InputEvent

twitchyliquid64:
I tested something like this:

```rust
pub extern "C" fn input_callback(event: *mut InputEvent, context: *mut c_void) {
    unsafe {
        core::mem::transmute::<*mut c_void, &MessageQueue<msg>>(context).put(msg((*event).clone()), 9999);
    }
}

fn main {
  // ...
  sys::view_port::input_callback_set(view_port, input_callback, core::mem::transmute(&queue));
  // ...
  
  loop {
      let evt = queue.get(9999999).unwrap().0;
      write!(&mut stdout, "Event: {}\r\n", evt).unwrap();
      stdout.flush().unwrap();
  
      if evt.key == keys::BACK {
          break;
      }
  }
  // ...
}
```

With output:

image

### DIFF
--- a/crates/sys/src/view_port.rs
+++ b/crates/sys/src/view_port.rs
@@ -1,13 +1,105 @@
 //! Low-level bindings to ViewPort API.
 
 use core::ffi::c_void;
+use core::fmt::Display;
 
 use crate::opaque;
 
 use super::canvas::Canvas;
 
 opaque!(ViewPort);
-opaque!(InputEvent);
+
+#[repr(C)]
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
I've gone ahead and applied your `repr(C)` vs `repr(transparent)` advice here as well, this one should be g2g (I hope).

dcoles (approved):

dcoles:
@twitchyliquid64 All good! Thanks for the PR!

PR #4, opened 2022-09-30
0.2-dev: Icon & canvas drawing bindings

twitchyliquid64:
Im sorry im on a roll today

### DIFF
--- a/crates/sys/src/canvas.rs
+++ b/crates/sys/src/canvas.rs
@@ -1,12 +1,116 @@
 //! Low-level bindings to the Canvas API.
 
 use core::ffi::c_char;
+use core::fmt::Display;
 
 use crate::opaque;
 
 opaque!(Canvas);
 
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Align(pub u8);
+
+impl Align {
+    /// Describes the alignment indicated.
+    pub fn description(self) -> &'static str {
+        use alignment::*;
+
+        match self {
+            LEFT => "Left",
+            RIGHT => "Right",
+            TOP => "Top",
+            BOTTOM => "Bottom",
+            CENTER => "Center",
+            _ => "Unknown",
+        }
+    }
+}
+
+impl Display for Align {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}", self.description())
+    }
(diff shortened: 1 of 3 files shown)


crates/sys/src/canvas.rs
```diff
@@ -1,12 +1,116 @@
 //! Low-level bindings to the Canvas API.
 
 use core::ffi::c_char;
+use core::fmt::Display;
 
 use crate::opaque;
 
 opaque!(Canvas);
 
+#[repr(C)]
```
dcoles, line 10: Use [`#[repr(transparent)]`](https://doc.rust-lang.org/reference/type-layout.html#the-transparent-representation) when creating C-compatible "new-types".

The issue with using `#[repr(C)]` in this specific case is that `Align` will have the layout of a C `struct` with a single `uint8_t`, as opposed to the layout of a bare `uint8_t`.

dcoles: In this case it would also be OK to use an `enum` (with `#[repr(C)]`).

The reason for creating a new-type for `Status` was that that it's undefined behaviour to cast an unknown `int` to a Rust `enum` and we have no control over what the C API might return. However when it's the other way around, we know the Rust values are always valid to pass to the C API.

crates/sys/src/icon.rs
```diff
@@ -0,0 +1,83 @@
...
+            pub static DOWN: Icon;
+
+            #[link_name = "I_ButtonLeft_4x7"]
+            pub static LEFT: Icon;
+            #[link_name = "I_ButtonRight_4x7"]
+            pub static RIGHT: Icon;
+        }
+    }
+
+    pub mod arrows {
```
dcoles, line 44: Maybe this one should be `pin`?

dcoles (changes requested):

crates/sys/src/canvas.rs
twitchyliquid64: Done!

> The reason for creating a new-type for Status was that that it's undefined behaviour to cast an unknown int to a Rust enum and we have no control over what the C API might return.

My usual approach for such enums is to implement `From<u32>`, but the downside is you have to implement the mappings from value to enum manually.

crates/sys/src/icon.rs
twitchyliquid64: Done.

dcoles (approved):

PR #5, opened 2022-10-01
Memory allocator

maybe-hello-world:
Added bindings to the furi memory allocator. 
Alloc crate should be usable now.

How to use: do nothing, #[global_allocator] will automatically wire it.
Comments:
- Tested on @twitchyliquid64 fz-rust example + vec/string usage.
- Current `on_oom` implementation is rather simple, feel free to propose what should be there or freely modify it. I'm not sure flipper would be able to properly allocate c_string on OOM (in tests I just see that somewhere furi_check fails and flipper reboots)

### DIFF
--- a/crates/flipperzero/src/alloc.rs
+++ b/crates/flipperzero/src/alloc.rs
@@ -1,0 +1,55 @@
+use core::alloc::{GlobalAlloc, Layout};
+use core::ffi::c_void;
+
+use flipperzero_sys::c_string;
+use flipperzero_sys::furi;
+
+extern "C" {
+    #[link_name = "free"]
+    pub fn free(p: *mut c_void);
+
+    #[link_name = "aligned_malloc"]
+    pub fn aligned_malloc(size: usize, align: usize) -> *mut *mut c_void;
+
+    #[link_name = "realloc"]
+    pub fn realloc(p: *mut c_void, size: usize) -> *mut c_void;
+}
+
+pub struct FuriAlloc;
+
+unsafe impl GlobalAlloc for FuriAlloc {
+    #[inline]
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        aligned_malloc(
+            layout.size(),
+            layout.align().max(core::mem::size_of::<usize>()),
+        ) as *mut u8
+    }
+
+    #[inline]
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        free(ptr as *mut c_void);
+    }
+
+    #[inline]
+    unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+        // https://github.com/flipperdevices/flipperzero-firmware/issues/1747#issuecomment-1253636552
+        self.alloc(layout)
+    }
+
+    #[inline]
+    unsafe fn realloc(&self, ptr: *mut u8, _layout: Layout, new_size: usize) -> *mut u8 {
+        realloc(ptr as *mut c_void, new_size) as *mut u8
+    }
+}
+
+#[global_allocator]
+static ALLOCATOR: FuriAlloc = FuriAlloc;
+
+#[alloc_error_handler]
+fn on_oom(_layout: Layout) -> ! {
+    unsafe {
+        furi::thread_yield();
+        furi::crash(c_string!("Rust: Out of Memory\r\n"))
+    }
+}
--- a/crates/flipperzero/src/alloc.rs
+++ b/crates/flipperzero/src/alloc.rs
@@ -5,13 +5,8 @@ use flipperzero_sys::c_string;
 use flipperzero_sys::furi;
 
 extern "C" {
-    #[link_name = "free"]
-    pub fn free(p: *mut c_void);
-
-    #[link_name = "aligned_malloc"]
-    pub fn aligned_malloc(size: usize, align: usize) -> *mut *mut c_void;
-
-    #[link_name = "realloc"]
+    pub fn aligned_free(p: *mut c_void);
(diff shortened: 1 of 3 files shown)


crates/flipperzero/src/alloc.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use flipperzero_sys::c_string;
+use flipperzero_sys::furi;
+
+extern "C" {
+    #[link_name = "free"]
+    pub fn free(p: *mut c_void);
+
+    #[link_name = "aligned_malloc"]
+    pub fn aligned_malloc(size: usize, align: usize) -> *mut *mut c_void;
```
dcoles, line 12: Why the double pointer to `void`? This doesn't match the definition in [`furi/core/memmgr.h`](https://github.com/flipperdevices/flipperzero-firmware/blob/0f9ea925d30f7a6183abe1002d680e3d1613a159/furi/core/memmgr.h#L45):

```c
void* aligned_malloc(size_t size, size_t alignment);
```

crates/flipperzero/src/alloc.rs
```diff
@@ -0,0 +1,55 @@
+use core::alloc::{GlobalAlloc, Layout};
+use core::ffi::c_void;
+
+use flipperzero_sys::c_string;
+use flipperzero_sys::furi;
+
+extern "C" {
+    #[link_name = "free"]
```
dcoles, line 8: Since the Rust function name and the symbol names are identical, using `#[link_name]` isn't strictly necessary.

crates/flipperzero/src/alloc.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+pub struct FuriAlloc;
+
+unsafe impl GlobalAlloc for FuriAlloc {
+    #[inline]
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        aligned_malloc(
+            layout.size(),
+            layout.align().max(core::mem::size_of::<usize>()),
```
dcoles, line 25: Why do we need to force a minimum alignment of `usize` (32-bits)?

crates/flipperzero/src/alloc.rs
```diff
@@ -0,0 +1,55 @@
...
+        ) as *mut u8
+    }
+
+    #[inline]
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        free(ptr as *mut c_void);
```
dcoles, line 31: Because `aligned_malloc` returns a pointer to the aligned byte, not the actual heap allocation, we have to pair it with [`aligned_free`](https://github.com/flipperdevices/flipperzero-firmware/blob/0f9ea925d30f7a6183abe1002d680e3d1613a159/furi/core/memmgr.c#L109).

crates/flipperzero/src/alloc.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[inline]
+    unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+        // https://github.com/flipperdevices/flipperzero-firmware/issues/1747#issuecomment-1253636552
+        self.alloc(layout)
+    }
+
+    #[inline]
+    unsafe fn realloc(&self, ptr: *mut u8, _layout: Layout, new_size: usize) -> *mut u8 {
+        realloc(ptr as *mut c_void, new_size) as *mut u8
```
dcoles, line 42: Unfortunately there isn't an `aligned_realloc`, thus we'll need to implement an aligned version of [`realloc`](https://github.com/flipperdevices/flipperzero-firmware/blob/0f9ea925d30f7a6183abe1002d680e3d1613a159/furi/core/memmgr.c#L20) ourselves.

dcoles (changes requested):
Hi @maybe-hello-world,

Thanks for the PR! Much appreciated!

I'd been putting off implementing a global allocator because of the `nightly` requirement, but if we want to use `alloc` there's not really any alternative until rust-lang/rust#66741 is implemented.

maybe-hello-world:
Sorry, I forgot to put [WIP] into the title, as this was just a draft implementation and intention to discuss.

I fixed mistakes and implemented suggestions, renamed alloc to furi_alloc to avoid ambiguities when using alloc crate, and added info functions about free heap and total heap.

I'm currently thinking about aligned_realloc.

maybe-hello-world:
@dcoles I think for now we can fallback to default rust GlobalAlloc realloc implementation (alloc + copy + free), even realloc in memmgr.c currently is implemented in that way. So, I dropped realloc implementation to let Rust generate the default implementation.

I think the code should be fine now.

dcoles:
> Sorry, I forgot to put [WIP] into the title, as this was just a draft implementation and intention to discuss.

No worries! Happy to see the interest!

Let me take a look over your changes.

crates/flipperzero/src/furi_alloc.rs
```diff
@@ -0,0 +1,57 @@
...
+use flipperzero_sys::c_string;
+use flipperzero_sys::furi;
+
+extern "C" {
+    fn aligned_free(p: *mut c_void);
+    fn aligned_malloc(size: usize, align: usize) -> *mut c_void;
+    fn memmgr_get_total_heap() -> usize;
+    fn memmgr_get_free_heap() -> usize;
+    fn memmgr_get_minimum_free_heap() -> usize;
+}
```
dcoles, line 13: I have some [bindings for `memmgr` functions on the `0.2-dev` branch](https://github.com/dcoles/flipperzero-rs/blob/0.2-dev/crates/sys/src/furi/memmgr.rs), but it's fine use these ones until it's merged.

dcoles (approved):
Thanks for the PR!

PR #6, opened 2022-10-01
0.2-dev: LED and vibration motor support

twitchyliquid64:
Use it like:

```rust
let notifications = sys::furi::record::open(RECORD_NOTIFICATION) as *mut Notification;

sys::notification::message(notifications, &sys::notification::led:red::BLINK);

sys::furi::record::close(RECORD_NOTIFICATION);
```

### DIFF
--- a/crates/sys/src/notification.rs
+++ b/crates/sys/src/notification.rs
@@ -1,0 +1,93 @@
+//! Low-level bindings to Notificcation API.
+
+use core::ffi::c_char;
+use crate::{opaque, c_string};
+
+pub const RECORD_NOTIFICATION: *const c_char = c_string!("notification");
+
+opaque!(Notification);
+opaque!(NotificationSequence);
+
+extern "C" {
+    #[link_name="notification_message"]
+    pub fn message(notifications_app: *mut Notification, sequence: *const NotificationSequence);
+}
+
+pub mod vibro {
+    use super::NotificationSequence;
+
+    extern {
+        #[link_name="sequence_set_vibro_on"]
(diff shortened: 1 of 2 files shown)


crates/sys/src/notification.rs
```diff
@@ -0,0 +1,95 @@
...
+        #[link_name="sequence_reset_vibro"]
+        pub static OFF: NotificationSequence;
+    }
+}
+
+pub mod led {
+    use super::NotificationSequence;
+
+    pub mod red {
+        use super::NotificationSequence;
```
dcoles, line 31: You might be able to avoid some of the `use super::NotificationSequence` by using `use super::super::*`. 🌟

crates/sys/src/notification.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        extern {
+            #[link_name="sequence_blink_red_10"]
+            pub static BLINK: NotificationSequence;
+            #[link_name="sequence_blink_red_100"]
+            pub static FLASH: NotificationSequence;
```
dcoles, line 37: I'm wondering if we should go with `BLINK_10` and `BLINK_100` to match the internal symbols (at least in the `-sys` crate).

crates/sys/src/notification.rs
twitchyliquid64: Done.

crates/sys/src/notification.rs
twitchyliquid64: Yeah that makes sense. Done.

dcoles:
Thanks again!

PR #12, opened 2022-10-18
A few more UI bindings

twitchyliquid64:
Any chance we could get a release after this? I have an app that I would like to release without patching the dep to a git repo :)

### DIFF
--- a/crates/sys/src/gui/mod.rs
+++ b/crates/sys/src/gui/mod.rs
@@ -6,6 +6,7 @@ use core::fmt::Display;
 
 pub mod canvas;
 pub mod elements;
+pub mod text_input;
 pub mod variable_item_list;
 pub mod view;
 pub mod view_dispatcher;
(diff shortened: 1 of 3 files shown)


dcoles:
@twitchyliquid64 I'm very close to merging this [`bindgen` branch|https://github.com/dcoles/flipperzero-rs/tree/bindgen] that switches out manually created bindings for ones automatically created from the Flipper Zero SDK (thus supporting everything all in one go).

The downside is that the bindings are not quite as nice as those done by hand, but with the size of the SDK, I realized that it wasn't going to be tenable to keep up with API changes or breakages.

dcoles:
The bindgen branch is now merged and published as [`0.3.0-alpha`](https://crates.io/crates/flipperzero/0.3.0-alpha).

Let me know if you spot any critical issues and I'll fix them before I push the official `0.3.0` crates.

twitchyliquid64:
I'll be on a plane tomorrow, lemme see if I can get it building with the new bindgen stuff.

dcoles:
Hi @twitchyliquid64. I'm going to close this PR since it hasn't had an update for a while, but we can always reopen it if you want to continue the work.

Thanks!

PR #16, opened 2022-12-04
Add Windows x86_64 and Linux x86 paths to generate-bindings

SitiSchu:
I've simply added the x86 path to Linux for consistency reasons.


dcoles (approved):
Thanks for the PR! Much appreciated!

PR #24, opened 2022-12-17
Add safe Dialog bindings

zhaofengli:
This PR adds safe dialog APIs which should make creating simple interactions easier. Currently, `DialogMessage` takes `&CStr`s so the caller can choose how to pass strings, but for ergonomics we can make `DialogMessage` own `CString`s and have APIs that automatically copy/convert from `&str`.


dcoles:
Hi @zhaofengli,

Thank you very much for the PR! This looks great! I'd be more than happy to accept this.

(Sorry for the slow response - the end of this year has been particularly crazy).

dcoles:
> Currently, `DialogMessage` takes `&CStr`s so the caller can choose how to pass strings, but for ergonomics we can make `DialogMessage` own `CString`s and have APIs that automatically copy/convert from `&str`.

So far I've erred on the side of caution—certain parts of the standard library can bloat binary size extremely quickly (e.g. `fmt`). Owned strings often get hit twice—rodata in the binary image and then when we place a copy on the heap.

I've been considering making `alloc` the default behaviour for `flipperzero` (if it doesn't hurt binary size when not using any allocating functions), but I can still imagine cases where one might want to forgo the provided allocator (perhaps wanting to use a slab or arena).

crates/flipperzero/src/furi/dialog.rs
```diff
@@ -0,0 +1,179 @@
...
+        assert!(!data.is_null());
+
+        Self {
+            data,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Sets the labels of the buttons.
+    pub fn set_buttons_c(
```
dcoles, line 68: I think we can drop the `_c` suffix (it took me a moment to realize what it meant). I don't think it's unreasonable to require a `CStr` if that's what the underlying API requires (even if `CStr`/`CString` are a bit of a pain to work with).

dcoles (approved):

PR #26, opened 2022-12-30
feat: allow trailing comas in function-like macro calls

JarvisCraft:
# Description

This allows using trailing comas in function-like macro calls.

This affects all public macros except for `entry!`
which generally looks like a good atrtibute-macro candidate.

### DIFF
--- a/crates/flipperzero/src/macros.rs
+++ b/crates/flipperzero/src/macros.rs
@@ -2,7 +2,7 @@
 
 #[macro_export]
 macro_rules! print {
-    ($pat:expr, $($args:tt)* $(,)?) => {{
+    ($pat:expr, $($args:tt)*) => {{
         $crate::furi::io::_print(core::format_args!($pat, $($args)*));
     }};
 
@@ -13,7 +13,7 @@ macro_rules! print {
 
 #[macro_export]
 macro_rules! println {
-    ($pat:expr, $($args:tt)* $(,)?) => {{
+    ($pat:expr, $($args:tt)*) => {{
         $crate::furi::io::_print(core::format_args!(concat!($pat, "\r\n"), $($args)*));
     }};
 
--- a/crates/flipperzero/src/macros.rs
+++ b/crates/flipperzero/src/macros.rs
@@ -2,22 +2,22 @@
 
 #[macro_export]
 macro_rules! print {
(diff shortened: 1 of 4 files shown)


crates/flipperzero/src/macros.rs
```diff
@@ -2,22 +2,22 @@
 
 #[macro_export]
 macro_rules! print {
-    ($pat:expr, $($args:tt)*) => {{
+    ($pat:expr, $($args:tt)* $(,)*) => {{
```
dcoles, line 5: The `print!` and `println!` macros  should already handle trailing commas. For example, the code below compiles without issue. I believe this works because `,` is a valid token, thus is matched by `$($args:tt)*`.

```rust
use flipperzero::{println, print};

print!("Hello {}", 42,);
println!("Hello, Rust! {}", 42,);
```

crates/rt/src/manifest.rs
```diff
@@ -23,7 +23,7 @@ const DEFAULT_STACK_SIZE: u16 = 2048; // 2 KiB
 /// ```
 #[macro_export]
 macro_rules! manifest {
-    ($($field:ident = $value:tt),*) => {
+    ($($field:ident = $value:expr),* $(,)*) => {
```
dcoles, line 26: I would suggest `$(,)?` to match the [standard libraries behaviour](https://sourcegraph.com/github.com/rust-lang/rust@ce85c98575e3016cf2007d90a85be321e592aa96/-/blob/library/core/src/macros/mod.rs?L554) of accepting at most one trailing comma.

dcoles (commented):
Hi @JarvisCraft,

Thank you for the pull request! Good spotting that these macros don't handle the trailing comma case. Do you mind switching `$(,)*` to `$(,)?` so the macros match the behaviour of functions to accept one trailing comma at most.

crates/flipperzero/src/macros.rs
JarvisCraft: You're right. I've removed optional trailing comas from macros ending with repeated `tt`s (i.e. `print!` and `println!`).

crates/rt/src/manifest.rs
JarvisCraft: My bad, I've pushed the commit resolving this issue.

dcoles (approved):

PR #27, opened 2022-12-31
fix(flipperzero-tools): fix conditional compilation on unix

JarvisCraft:
# Description

This fixes the typo in the name of cfg-parameter responsible for choosing the used OS.

This also updates dependencies to their latest versions
and formats `generate-bindings.rs` applying rustfmt's and clippy's fixes.


dcoles (approved):
Thanks for the fix and updating the build dependencies.

PR #28, opened 2022-12-31
build: fix build of `crates` inIntellij-based IDEs

JarvisCraft:
# Description

This fixes build of `crates` module in Intellij-based IDEs.

Previous configuration would cause errors while trying to configure Cargo-based project.


dcoles (approved):
Interesting. That [should be allowed](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget), but not an issue to switch to the "string" form.

PR #29, opened 2022-12-31
feat: implement wrapper for GUI APIs

JarvisCraft:
# Description

This is a WIP PR to provide generic safe bindings for Flipper's GUI APIs.


crates/gui/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "flipperzero-gui"
```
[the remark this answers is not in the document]
JarvisCraft: Yup, at the moment I am not that sure about the placement thus I use `gui` as some form of incubator for the new APIs, but I already see this separation being a problem (as can be seen by a comment about `Align` type in this `Cargo.toml`). In the end I am planning to either move all this APIs to `flipperzero` (probably behind feature-flags) all provide more logical separation.

crates/gui/src/view_port.rs
```diff
@@ -0,0 +1,389 @@
...
+    ///
+    /// let view_port = ViewPort::new();
+    /// let width = view_port.get_width();
+    /// ```
+    pub fn get_width(&self) -> NonZeroU8 {
+        // SAFETY: `self.view_port` is always valid
+        unsafe { sys::view_port_get_width(self.view_port) }
+            .try_into()
+            .expect("`view_port_get_width` should produce a positive value")
+    }
```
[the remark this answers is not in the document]
JarvisCraft: Hmm, had to check the sources and it seems that no zero-check is done on `view_port_set_width` (i.e. callers have to treat zero as a default themselves). I will update the API to no unwrap the Result.

JarvisCraft:
By the way, while the API is in its early stage, I would like to hear general thoughts from you if I am working in the right direction.

dcoles:
So far this is looking great. I think everyone will appreciate not having to use `sys` to put something on the screen.

One issue I did run into in the past was how to implement GUI callbacks. It would be really nice to use Rust closures, but closures require a "fat pointer" which is tricky to use with the standard `void* context` API (Perhaps we say closures require `alloc::Box` and leave it at that).

> I am trying my best to make the API as safe as possible but the chances are high that the initial implementation will contain unsound bugs. Thus intensive review will be much appreciated.

So far everything looks good. I think even things like [`as_ptr`](https://doc.rust-lang.org/stable/alloc/vec/struct.Vec.html#method.as_ptr) are safe (so long as you never dereference the pointer).

The one I'm not 100% sure about is [`NonNull`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html), because it implies the type is [covariant](https://doc.rust-lang.org/reference/subtyping.html#variance). I *think* it's OK (`&'static ViewPort` would be a subtype of `&'a ViewPort`), but I don't understand variance all that well.

JarvisCraft:
> One issue I did run into in the past was how to implement GUI callbacks.

As for now, I think about one way to implement this: theoretically, we could provide dispatcher-functions matching the C-signature and pass the actual closures via context. The dispatch function would simply deref context calling its actual function.

Ir still is just a concept with which I am going to experiment (at least, I expect some `Pin` trickery).

> I think even things like as_ptr are safe (so long as you never dereference the pointer).

Seems true, as for now I marked them as unsafe (partially, to make myself think twice before using them), but it seems that they really aren't unsafe and ATM the safety contract I write for `as_ref()` matches that of the following usage of the pointer, so, most probably, I will safen them.

> But I don't understand variance all that well.

Same problem :D I will have to re-read documentation about it once the API is closer to its final form.

---

Thanks for your review and the comments! And have a happy new year by the way :rocket:

dcoles:
> As for now, I think about one way to implement this: theoretically, we could provide dispatcher-functions matching the C-signature and pass the actual closures via context. The dispatcher-function would simply deref context calling its actual function.

That's where I ran into an issue:

```rust
pub fn set_draw_callback(&self, callback: &dyn Fn(&Canvas)) {
    // this won't compile because callback is a fat pointer (i.e. `sizeof(&dyn Fn()) > sizeof(*void)`)
    let callback = callback as *mut c_void;

    unsafe {
        sys::view_port::draw_callback_set(self.inner, Self::draw_callback, callback);
    }
}
```

Without a fat pointer, Rust wouldn't know the correct vtable for the concrete trait implementation.

JarvisCraft:
`&dyn Fn(&Canvas)` may be stored in some wrapper struct so that we would have to only have the word-sized pointer to the structure. Something similar to:

```rust
struct CallbackWrapper<'a>(&'a dyn Fn(&Canvas));

impl<'a> CallbackWrapper {
    fn call(&self, canvas: &Canvas) {
        self.0(canvas) // vcall ocurrs here
    } 
}
```

JarvisCraft:
Here are some good news: `examples/gui` is now is now `#![forbid(unsafe_code)]` and also has additional functionality:
- it now uses queue to exit on `BACK` key;
- it now has a timer controlled by buttons.

JarvisCraft:
Here are some good news: `examples/gui` is now is now `#![forbid(unsafe_code)]` and also has additional functionality:
- it now uses queue to exit on `BACK` key;
- it now has a timer controlled by buttons.

[Demo](https://user-images.githubusercontent.com/7693005/210294672-7a3b7901-ace4-4310-9910-cd74b980ca1e.webm)

JarvisCraft:
Another update here with enhanced XBM support:

image

JarvisCraft:
Hi, @dcoles! Since you've added `gui` as an explicit module ni your recent commits, should I migrate current code to it?

dcoles:
> Hi, @dcoles! Since you've added `gui` as an explicit module ni your recent commits, should I migrate current code to it?
 
Yes please. I'm trying to roughly maintain the same structure as in the Flipper Zero SDK.

This would be particularly useful if we want to put various modules under a feature flag.

JarvisCraft:
As a small status update on this PR: I'm out for my holidays thus I'm taking some rest from development (also, I don't have my flipper with me for testing my changes), but I will get back to work as soon as I come back :)

dcoles:
> As a small status update on this PR: I'm out for my holidays thus I'm taking some rest from development (also, I don't have my flipper with me for testing my changes), but I will get back to work as soon as I come back :)

Awesome! Have a great break! 🏝️

tocklime:
> One issue I did run into in the past was how to implement GUI callbacks. It would be really nice to use Rust closures, but closures require a "fat pointer" which is tricky to use with the standard `void* context` API (Perhaps we say closures require `alloc::Box` and leave it at that).

I think this is doable using the trick detailed [here ('improving ergonomics' section)](https://users.rust-lang.org/t/callback-based-c-ffi/26583/5) - I haven't tested this much, and I'm sure it's not sound, but something like this seems to be working, at least on a trivial example:

```rust
    unsafe extern "C" fn wrapped_callback<Env>(
        canvas: *mut sys::Canvas,
        context: *mut core::ffi::c_void,
    ) where
        Env: Sized + FnMut(&Canvas),
    {
        unsafe {
            let f_raw_ptr: *mut Env = context.cast();
            let f = f_raw_ptr.as_mut().expect("Null fptr");
            let canvas = Canvas { inner: canvas };
            f(&canvas);
        }
    }

    pub fn draw_callback_set<F>(&self, callback: &F)
    where
        F: FnMut(&Canvas),
    {
        unsafe {
            sys::view_port_draw_callback_set(
                self.inner,
                Some(Self::wrapped_callback::<F>),
                callback as *const F as *mut core::ffi::c_void,
            );
        }
    }

```

which is used something like u
```rust
    let hi = core::ffi::CStr::from_bytes_with_nul(b"Hi There!\0").unwrap();

    view_port.draw_callback_set(&(|c: &Canvas| {
        c.draw_str(39, 31, hi);
    }));
```

tocklime:
You're right, they do, of course. I was experimenting with closure-based callbacks, ideally so the calling code can be something like `view_port.on_draw(|canvas| { /*...do things with canvas... */ };`, rather than using `ViewPortCallbacks`.

JarvisCraft:
It can be replaced with singular callbacks (in fact, current Callbacks structure is not that different).
Moreover, we can add a struct implementing `ViewPortCallbacks` which would allow specifying callbacks via its fields, a la

```rust
struct MutViewPortCallbacks {
    draw_callback: Box<dyn Fn(&mut CanvasView)>,
}

impl ViewPortCallbacks for MutViewPortCallbacks {
    fn on_draw(&mut self, canvas: & CanvasView) {
        self.draw_callback(canvas);
    } 
}

impl MutViewPortCallbacks {
    fn set_on_draw(&mut self, callback: impl Fn(&mut CanvasView) {
        self.draw_callback = Box::new(callback);
    }
}
```

It would only require adding a method to `ViewPort` for reading its `ViewPortCallback` so that its closures can be hot-swapped.

dcoles:
Hi @tocklime,

> I think this is doable using the trick detailed [here ('improving ergonomics' section)](https://users.rust-lang.org/t/callback-based-c-ffi/26583/5) - I haven't tested this much, and I'm sure it's not sound, but something like this seems to be working, at least on a trivial example:

That's pretty clever. I never thought of using generics for a `extern "C"` function, but it's not really much different than if you had written it yourself.

At a quick glance, it seems to be sound (I wrote a [cut-down example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=885551060b9c6bf58416aa9b6a6b418c)), but it does still need somewhere to store the actual callback (the `set_c_callback` only stores a pointer to some data, not the actual data itself).

So you'd still need to store the callback environment in either `main` or using on the heap using `Box`. But it still potentially saves you one level of `Box`-ing (i.e. no need for a `Box<dyn Fn(&Canvas)>`).

JarvisCraft:
> So you'd still need to store the callback environment in either main or using on the heap using Box.

Using stack location is problematic since we would need a way to enforce the immovability of the pointer (pinning of it).

As can be seen [here](https://github.com/JarvisCraft/flipperzero-rs/blob/333e30e456523c148315e499694008eac9f099c5/crates/flipperzero/src/gui/view_port.rs#L42-L66), generic extern-C function is already used for this purpose. Specific callback instance is stored in a `Box`, there are no vcalls involved and the only level of indirection is the need for dispatch function which seems to be obligatory.

JarvisCraft:
@dcoles, @str4d hi there! Since most of the other PRs involving other APIs have been merged successfully, I guess it is the moment we get back to the discussion of the best API for calbacks in GUIs (see the prvious message).

dcoles:
Hey @JarvisCraft. Sorry for the delay in responding. I'm traveling at the moment, so I've been limited on time in front of a computer.

I think this is a good solution that balances ergonomics with efficiency. I'm happy if you want to move forward with this approach.

str4d:
My without-review inclination is to merge this PR as-is and make further improvements in subsequent PRs. I'm going to attempt to review the PR sometime in the next week or two, and see if that changes my opinion.

@JarvisCraft there are merge conflicts that need resolving before the PR can move forward.

str4d:
Re: the callback approach, I think we should work with what is implemented now, and then a subsequent PR could completely change how the callbacks are implemented. We don't need to consider the current approach final, and in some respects it would be better to get some experience with using it, to figure out how many downsides we run into in practice.

JarvisCraft:
> I think this is a good solution that balances ergonomics with efficiency. I'm happy if you want to move forward with this approach.

@dcoles, Thanks for your review! in this case, I will stick to this approach.

> My without-review inclination is to merge this PR as-is and make further improvements in subsequent PRs. I'm going to attempt to review the PR sometime in the next week or two, and see if that changes my opinion.

@str4d, good point. I like the idea of merging this PR as-is (ensuring compilabilty and test&example-stability, of cource) in order to provide "beta" version of this wrappers (hidden behind a feature flag such as `unstable-service-gui`) and then iteratively upgrade it to finally fully and soundly cover the APIs.

> @JarvisCraft there are merge conflicts that need resolving before the PR can move forward.

Yup, I will try to resolve them throught this week.

JarvisCraft:
@dcoles, @str4d, I guess it's time you start reviewing this monstrocity of a PR

# Highlights

* A log of `gui` service APIs are implemented although many more are to come in the future PRs.

* Callback approach has been unchanged since the previous highlights.
  * Mutable callbacks have not yet been implemented since we will have to double-check their soundness and they don't seem to be a feature with which we should rush.

* All APIs for `service`s have been hidden under the respective feature flags which has been suggested [here](https://github.com/flipperzero-rs/flipperzero/pull/29#discussion_r1059691569) by @dcoles.
  * This has broken the tests machinery since we now need to enable `--all-features` for it. I am not sure what would be the right way to do this (maybe just hide tests in `tests/`) this @str4d's suggestions would great.

* Some of the existing examples have been rewritten using the new APIs making `messagess` API almost 100%-safe (with the only exception being immage creation for which we may implement a proc-macro in the future).

* An API has been added for XBM images including basic image manipulation and an ability to safely compile-time embed them.
  * There is also a decl-macro `xbm!` allowing embedding of XBM-formatted images into a macro (usage can be [seen](https://github.com/JarvisCraft/flipperzero-rs/blob/gui-bindings/crates/flipperzero/examples/xbms/ferris.rs) in the `gui` example). It was more of a proof-of-concept (especially a trick to validate names in a decl-macro) and will probably be replaced with a proc-macro in the future (also allowing loading from external files such as by `include_str!`).

* I've enhanced documentation:
  * Enabled automatic generation of feature information:
    image
  * Enabled generation of feature-gate info on items:
   image
  * Fixed a few broken docs (specifically, invalid links) in other items which were reported by `cargo doc` (it's worth mentioning that references to crypto-APIs are still broken as their dependencies are dev-only; also doxygen creates a ton of broken links):
  image

* Some existing APIs were minorly changed to fit better (e.g. `UnsafeRecord::as_ptr` has been renamed to `UnsafeRecord#as_raw` so that the name i smore universal).

* I've made `flipperzero_alloc` re-export of items from `alloc` so that users can only extern the former without a need to extern the latter (this has been reflected in the examples involving `alloc` feature).

# Soundness notes

As mentioned before, I do have some concerns about soundness. Especially, I stil am not that sure if current callbacks correclt yprovide `&mut`-access. I guess that it it fone to go with the current approach as a starting point and explore this aspect in the future Issues/PRs.

JarvisCraft:
By the way, I suggest that we use root crate (the on with workspaces) to host `flipperzero` crate since this would simplify project structure a bit also allowing us to configure `[package.metadata.docs.rs]` which for some reason does not work with workspaces when specified at crate-level.

str4d:
> By the way, I suggest that we use root crate (the on with workspaces) to host `flipperzero` crate since this would simplify project structure a bit also allowing us to configure `[package.metadata.docs.rs]` which for some reason does not work with workspaces when specified at crate-level.

I disagree with this suggestion. It complicates the project structure (no longer simply flat) and can cause issues with `cargo` not knowing what you want run (IIRC meaning you basically have to pass the `--workspace` flag to every `cargo` invocation).

`[package.metadata.docs.rs]` doesn't work in a virtual root, but it should work in each sub-crate fine (I've done this in other projects). `cargo` also supports the `workspace.metadata` key (https://github.com/rust-lang/cargo/issues/8309), so if docs.rs doesn't listen to that for common configs across the workspace, we should open an issue upstream.

str4d:
Re: this PR, the summary arms to indicate there are a lot of changes that are maybe used by the GUI changes, but are themselves unrelated to the GUI. To keep the complexity manageable, I would prefer that we review those in separate smaller PRs, and keep this PR for the (still substantial) GUI changes (rebasing as necessary when the smaller PRs are merged).

JarvisCraft:
> [package.metadata.docs.rs] doesn't work in a virtual root, but it should work in each sub-crate fine (I've done this in other projects). cargo also supports the workspace.metadata key (https://github.com/rust-lang/cargo/issues/8309), so if docs.rs doesn't listen to that for common configs across the workspace, we should open an issue upstream.

Per-crate approach failed to work, at least for things like `rustdoc-args` and `cargo-args` but I will try `workspace.metadata` approach.

> I would prefer that we review those in separate smaller PRs, and keep this PR for the (still substantial) GUI changes (rebasing as necessary when the smaller PRs are merged).

Fair point, I will try to extract the parts unrelated to GUI APIs (documentation fixes and enhancments, feature-flags, minor refactoring) to separate PRs. As for now the essential part of this PR may be reviewed at [`flipperzero/src/gui`].

JarvisCraft:
First separated PR is ready: #81

crates/flipperzero/src/gui/mod.rs
```diff
@@ -1,3 +1,230 @@
...
+/// System Gui wrapper.
+pub struct Gui {
+    raw: UnsafeRecord<SysGui>,
+}
+
+impl Gui {
+    /// Furi record corresponding to GUI.
+    pub const RECORD: *const c_char = sys::c_string!("gui");
+
+    pub fn new() -> Self {
```
str4d, line 35: These new public APIs should all be documented. That would also help with review, knowing the intent behind each API (particularly the ones that aren't just a type-safe wrapper around a single SDK function).

crates/flipperzero/src/gui/mod.rs
```diff
@@ -1,3 +1,230 @@
...
+    ffi::c_char,
+    fmt::Debug,
+    ops::{Deref, DerefMut},
+};
+use flipperzero_sys::{
+    self as sys, furi::UnsafeRecord, Canvas as SysCanvas, Gui as SysGui, GuiLayer as SysGuiLayer,
+};
+
+/// System Gui wrapper.
+pub struct Gui {
```
str4d, line 27: This new struct should be mentioned in the changelog in the `Added` section. See prior changelog entries for the style.

crates/flipperzero/src/gui/mod.rs
```diff
@@ -1,3 +1,230 @@
 //! GUI service.
 
 pub mod canvas;
+pub mod icon;
+pub mod icon_animation;
+pub mod view;
+pub mod view_port;
+pub mod xbm;
```
str4d, line 8: These new modules should be mentioned in the changelog under the `Added` section.

crates/flipperzero/src/gui/mod.rs
```diff
@@ -1,3 +1,230 @@
...
+    raw: UnsafeRecord<SysGui>,
+}
+
+impl Gui {
+    /// Furi record corresponding to GUI.
+    pub const RECORD: *const c_char = sys::c_string!("gui");
+
+    pub fn new() -> Self {
+        // SAFETY: `RECORD` is a constant
+        let gui = unsafe { UnsafeRecord::open(Self::RECORD) };
```
str4d, line 37: It occurs to me that records are probably not supposed to be opened by multiple threads simultaneously. If that is the case, then separately from this PR, we might end up defining the SDK-provided records in some common location, where we can control access via a mutex. An alternative approach would be to create a harness for defining "record singletons", which would store them as heap statics (#64); then each service module would use it to define its record type.

The context for the above thought was "is it okay to call `Gui::new` twice?". Nothing in the current API will prevent it, but if the backing records are changed to be mutex-gated, then `Gui::new` would necessarily also be. And if we instead provide a record singleton harness, then `Gui::new` would become a private method.

crates/flipperzero/src/gui/mod.rs
```diff
@@ -1,3 +1,230 @@
...
+
+    pub fn as_raw(&self) -> *mut SysGui {
+        self.raw.as_raw()
+    }
+
+    pub fn add_view_port<VPC: ViewPortCallbacks>(
+        &mut self,
+        view_port: ViewPort<VPC>,
+        layer: GuiLayer,
+    ) -> GuiViewPort<'_, VPC> {
```
str4d, line 50: I don't see any examples with multiple view ports. Is that a thing we should expect to eventually support?
- If so, then storing `&self` inside `GuiViewPort` currently prevents this (because then we can't re-take a mutable reference to `self`), so we might need an alternative API for adding multiple view ports at once (or an alternative way to reference the parent instead of `&self`, but that still binds to the lifetime of `Self` rather than the `&mut self` or `&self`).
- If not, then `Gui::add_view_port` feels misnamed, and it should instead be something like `Gui::connect_view_port`.

crates/flipperzero/src/gui/mod.rs
```diff
@@ -1,3 +1,230 @@
...
+        unsafe { sys::gui_get_framebuffer_size(raw) }
+    }
+
+    pub fn set_lockdown(&self, lockdown: bool) {
+        let raw = self.raw.as_raw();
+        // SAFETY: `raw` is always a valid pointer
+        unsafe { sys::gui_set_lockdown(raw, lockdown) }
+    }
+
+    // TODO: separate `GuiCanvas` (locking the parent)
```
str4d, line 76: Has this been renamed?
```suggestion
    // TODO: separate `ExclusiveCanvas` (locking the parent)
```

str4d (commented):
Flushing an initial partial review of `flipperzero::gui`.

JarvisCraft:
By the way, @str4d, is there any way we could add suport for running MIRI?

Asking this since the whole callback magic is quite tricky from aliasing standpoint. I still am unsure which invariants should be upheld for `*mut T`s originating from `Box<T>` so it would be quite helpful.

I tried running it but the MIRId app just crashed.

crates/flipperzero/src/gui/view_port/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+    ///
+    /// Resize `ViewPort` to automatically selected dimensions:
+    ///
+    /// ```
+    /// use flipperzero::gui::view_port::ViewPort;
+    ///
+    /// let mut view_port = ViewPort::new(());
+    /// view_port.set_dimensions(None);
+    /// ```
+    pub fn set_dimensions(&mut self, dimensions: Option<(NonZeroU8, NonZeroU8)>) {
```
dcoles, line 233: Are there times you might want to `set_dimensions(Some(NonZeroU8), None)`?

crates/flipperzero/src/input/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+/// assert_eq!(0b10, sequence.source());
+/// assert_eq!(0b10101010_11110000_11111111, sequence.counter());
+/// ```
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug, uDebug, Hash, Eq, PartialEq, PartialOrd, Ord)]
+pub struct InputEventSequence(u32);
+
+impl InputEventSequence {
+    pub const fn source(self) -> u8 {
+        ((self.0 >> 30) & 0b11) as u8
```
dcoles, line 93: It might be nice to define these as `SOURCE_MASK` and `SOURCE_SHIFT` constants.

crates/flipperzero/src/internals.rs
```diff
@@ -0,0 +1,99 @@
...
+/// falling back to ad-hoc implementations otherwise.
+#[allow(dead_code)] // this functions may be unused if a specific feature set does not require them
+pub(crate) mod ops {
+    pub(crate) const fn div_ceil_u8(divident: u8, divisor: u8) -> u8 {
+        #[cfg(feature = "unstable_intrinsics")]
+        {
+            divident.div_ceil(divisor)
+        }
+        #[cfg(not(feature = "unstable_intrinsics"))]
+        {
```
dcoles, line 77: I think you can use `if cfg!`/`else` here:

```rust
if cfg!(feature = "unstable_intrinsics") {
    divident.div_ceil(divisor)
} else {
    let quotient = divident / divisor;
    let remainder = divident % divisor;

    if remainder > 0 && divisor > 0 {
        quotient + 1
    } else {
        quotient
    }
}
```

crates/flipperzero/src/internals.rs
```diff
@@ -0,0 +1,99 @@
...
+                quotient + 1
+            } else {
+                quotient
+            }
+        }
+    }
+}
+
+pub(crate) mod macros {
+    macro_rules! impl_std_error {
```
dcoles, line 90: Can you provide a short bit of documentation what this macro is for and when folks should use it.

crates/flipperzero/src/kernel.rs
```diff
@@ -0,0 +1,76 @@
...
+            1 => LockGuard {
+                was_locked: true,
+                _marker: PhantomData,
+            },
+            status => Err(LockError::ErrorStatus(Status(status)))?,
+        })
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
```
dcoles, line 35: Does `PartialEq`/`Eq` make sense on this? Would you ever do a comparison?

crates/flipperzero/src/kernel.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/// A type of error which can be returned whenever a lock is acquired.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub enum LockError {
+    Interrupted,
+    ErrorStatus(Status),
+}
+
+impl Display for LockError {
```
dcoles, line 67: Should we also implement `uDisplay`?

crates/flipperzero/src/xbm.rs
```diff
@@ -0,0 +1,479 @@
...
+                    (None, None) => true,
+                    _ => false,
+                }
+            }
+
+            let width_ident = stringify!($width_ident).as_bytes();
+            assert!(
+                matches!(width_ident, [.., b'_', b'w', b'i', b'd', b't', b'h']),
+                "width identifier should end with `_width`",
+            );
```
dcoles, line 299: I wonder if we should replace these with [`static_assertions`](https://crates.io/crates/static_assertions) (not necessarily in this PR).

I think the asserts make sense, though I do worry about the runtime/code-size overhead that we might be taking on. Probably worth doing a side-by-side with/without those asserts and see how much additional code they introduce.

Unfortunately binary size has a very direct impact on how quickly Flipper applications start.

dcoles (changes requested):
You weren't kidding about this being a big code review!

That said, it's shaping up really nicely. It's really impressive to see that it's now possible to write a full graphical application without the user needing to write any unsafe code. Looking forward to merging it in. 🙌

crates/flipperzero/src/xbm.rs
JarvisCraft: Generally, this would normally be replaced with a proc-macro in the future.

As for now, I think that it should normally be remomved by the compiler, considering that it is intended for use in `const`s where the expression has to be CT-evaluated.

crates/flipperzero/examples/gui.rs
```diff
@@ -4,53 +4,124 @@
...
-}
+const PLUS_IMAGE: XbmImage<ByteArray<8>> = XbmImage::new_from_array::<8, 8>([
+    0b00_11_11_00,
+    0b00_11_11_00,
+    0b11_11_11_11,
+    0b11_11_11_11,
+    0b11_11_11_11,
+    0b11_11_11_11,
+    0b00_11_11_00,
+    0b10_11_11_01,
```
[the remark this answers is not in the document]
JarvisCraft: Added a comment

crates/flipperzero/examples/gui.rs
```diff
@@ -4,53 +4,124 @@
...
+    let mut gui = Gui::new();
+    let mut view_port = gui.add_view_port(view_port, GuiLayer::Fullscreen);
+
+    let status = loop {
+        match exit_event_queue.get(Duration::from_millis(100)) {
+            Ok(()) => {
+                println!("Exit pressed");
+                break 0;
+            }
+            Err(e) => {
```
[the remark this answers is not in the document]
JarvisCraft: Done

crates/flipperzero/examples/xbms/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod ferris;
```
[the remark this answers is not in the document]
JarvisCraft: Renamed to the latter

crates/flipperzero/src/gui/canvas/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// ```
+    pub fn commit(&mut self) {
+        let raw = self.raw.as_ptr();
+        // SAFETY: `raw` is always valid
+        unsafe { sys::canvas_commit(raw) };
+    }
+
+    // FIXME:
+    //  - canvas_reset
+    //  - canvas_commit
```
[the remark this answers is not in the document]
JarvisCraft: Right, removed them

crates/flipperzero/src/input/mod.rs
JarvisCraft: All done

crates/flipperzero/src/internals.rs
JarvisCraft: Done

crates/flipperzero/src/kernel.rs
JarvisCraft: Yup, added

JarvisCraft:
Just found the explicit information about aliasing rules for `Box` which seem to be more restrictive as I previously thought: https://doc.rust-lang.org/std/boxed/index.html#considerations-for-unsafe-code

Another point on allowing MIRI runs on flipper.

cc @mogenson, this may be important for what you currently doing with `ViewPort`, although it is worth to just wait for me to write the right context- implementation here

crates/flipperzero/src/gui/view_dispatcher/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        // SAFETY: both pointers are guaranteed to be non-null
+        let view_dispatcher = {
+            let raw = unsafe { NonNull::new_unchecked(raw) };
+            Self { raw, callbacks }
+        };
+
+        if QUEUE {
+            // SAFETY: `raw` is a valid pointer
+            // and corresponds to a newly created `ViewPort`
```
mogenson, line 44: nit: `ViewDispatcher` instead of `ViewPort`

crates/flipperzero/src/gui/view_dispatcher/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+            unsafe { sys::view_dispatcher_set_event_callback_context(raw, context) };
+        }
+
+        if register_custom_event {
+            pub unsafe extern "C" fn dispatch_custom<C: ViewDispatcherCallbacks>(
+                context: *mut c_void,
+                event: u32,
+            ) -> bool {
+                let context: *mut C = context.cast();
+                // SAFETY: `context` is stored in a `Box` which is a member of `ViewPort`
```
mogenson, line 63: nit: `ViewDispatcher` instead of `ViewPort`

crates/flipperzero/src/gui/view_dispatcher/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+            // SAFETY: `raw` is valid
+            // and `callbacks` is valid and lives with this struct
+            unsafe { sys::view_dispatcher_set_custom_event_callback(raw, callback) };
+        }
+        if register_navigation_callback {
+            pub unsafe extern "C" fn dispatch_navigation<C: ViewDispatcherCallbacks>(
+                context: *mut c_void,
+            ) -> bool {
+                let context: *mut C = context.cast();
+                // SAFETY: `context` is stored in a `Box` which is a member of `ViewPort`
```
mogenson, line 78: nit: `ViewDispatcher` instead of `ViewPort`

crates/flipperzero/src/gui/view_dispatcher/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+            // SAFETY: `raw` is valid
+            // and `callbacks` is valid and lives with this struct
+            unsafe { sys::view_dispatcher_set_navigation_event_callback(raw, callback) };
+        }
+        if let Some(tick_period) = tick_period {
+            pub unsafe extern "C" fn dispatch_tick<C: ViewDispatcherCallbacks>(
+                context: *mut c_void,
+            ) {
+                let context: *mut C = context.cast();
+                // SAFETY: `context` is stored in a `Box` which is a member of `ViewPort`
```
mogenson, line 93: nit: `ViewDispatcher` instead of `ViewPort`

crates/flipperzero/src/gui/view_dispatcher/mod.rs
```diff
@@ -0,0 +1,172 @@
+use crate::internals::alloc::BoxNonNull;
+use alloc::boxed::Box;
+use core::num::NonZeroU32;
+use core::{
+    ffi::c_void,
+    ptr::{self, NonNull},
+};
+use flipperzero_sys::{self as sys, ViewDispatcher as SysViewDispatcher};
+
+pub struct ViewDispatcher<C: ViewDispatcherCallbacks, const QUEUE: bool> {
```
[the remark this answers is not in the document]
JarvisCraft: As for now, it is mostly my experiments with what we can and what we can't express with it.

My point of having `queue` as a const generic is to prohibit calls to methods requiring it at the type-system level. 

PS: to be fair, I've actually commmited my `ViewDispatcheer` experiments by accident, and then just went with experimenting on them

JarvisCraft:
@dcoles, coudld you please bump the API version? There is a number of changes in there (specifically, `ViewHolder` is now part of GUI services) and it is important for local testing

JarvisCraft:
# Status update

Currently trying to understand the right mental model of aliasing with respect to Flipper's [GUI lifecycle](https://miro.com/app/board/uXjVOa4cno8=/).

Since there is high risk of violating Rust's aliasing guarantees, there may be a need to use `UnsafeCell` (or its derivative *somewhere*), although I still am not that certain about it.

crates/flipperzero/src/gui/view_port/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+impl Drop for ViewPortInner {
+    fn drop(&mut self) {
+        let raw = self.0.as_ptr();
+        // SAFETY: `raw` is a valid pointer
+        unsafe { sys::view_port_free(raw) };
+    }
+}
+
+#[allow(unused_variables)]
+pub trait ViewPortCallbacks {
```
loftyinclination: This needs to be `Send` -- `on_draw` is invoked on the gui_srv thread, in response to a flag being set by `gui_update`.

crates/flipperzero/src/gui/view_dispatcher/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+    /// let mut view_dispatcher = ViewDispatcher::new(MyCallbacks {
+    ///     value: 10
+    /// }, &mut gui, ViewDispatcherType::Fullscreen);
+    ///
+    /// view_dispatcher.send_custom_event(20);
+    /// // should print `10 + 20 = 30`
+    /// ```
+    pub fn new(callbacks: C, gui: &'a Gui, kind: ViewDispatcherType) -> Self {
+        // discover which callbacks should be registered
+        let register_custom_event = !ptr::eq(
```
loftyinclination: I dont believe that this pointer equality check will ever pass. The vtable for () is different from the vtable for the default implementation, bcs of the override on tick_period, which generates an entirely ne table for the unit type. 

(Fixing this is tricky -- I think you might be able to do something with associated types on the trait, since pointer comparison cant be trusted.)

PR #44, opened 2023-02-24
chore: use workspace features more intensively in crates

JarvisCraft:
# Description

This PR:

- enhances the use of new Cargo Workspaces features in `crates/`;
- configures Cargo Workspaces for `examples/` directory.

# Limitations

Currently, due to a Cargo bug/design, when building with `cargo build` in a workspace directory, feature unification happens across all packages causing all of them to have `alloc` feature enabled thus leading to a compilation failure on crates which don't configure the allocator.

This **does not** happen when building packages independently (i.e. `cargo build -p foo` or `cd foo && cargo build`).

I consider this acceptable as for now since examples are already built one-by-one, although I will report this problem to Cargo Issues so that it can be resolved in the future.

More on [cargo zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60cargo.20build.60.20forced.20dependency.20on.20alloc/near/332428810) and [cargo issue](https://github.com/rust-lang/cargo/issues/4463).


crates/Cargo.toml
```diff
@@ -7,6 +7,19 @@ members = [
 ]
 resolver = "2"
 
+[workspace.package]
+version = "0.7.0"
```
dcoles, line 11: I'm expecting these package versions to start to diverge (at least for `flipperzero-sys`), but this is fine for now.

dcoles:
> Currently, due to a Cargo bug/design, when building with cargo build in a workspace directory, feature unification happens across all packages causing all of them to have alloc feature enabled thus leading to a compilation failure on crates which don't configure the allocator.

That's a particularly annoying bug. Unfortunately, I can see this constantly being reported as a bug if `cargo build` (and family) don't work correctly in the `examples` directory.

This might be good enough a reason to move all the examples into `crates/flipperzero/examples` and then create a dedicated template project for people to use as a reference.

In the past we've talked about whether we should just always enable `alloc` for the `flipperzero` crate. From a quick experiment, it looks like just enabling the `alloc` feature and defining a `#[global_allocator]` does not have any noticeable effect on the binary size. The main annoyance is that every binary needs to explicitly `extern crate flipperzero_alloc`, even if not otherwise used.

examples/dialog/Cargo.toml
```diff
@@ -17,7 +19,7 @@ bench = false
 test = false
 
 [dependencies]
-flipperzero = { version = "0.7.0", path = "../../crates/flipperzero", features = ["alloc"] }
-flipperzero-alloc = { version = "0.7.0", path = "../../crates/alloc" }
-flipperzero-sys = { version = "0.7.0", path = "../../crates/sys" }
-flipperzero-rt = { version = "0.7.0", path = "../../crates/rt" }
```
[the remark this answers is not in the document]
JarvisCraft: Yup, this sounds reasonable. But I feel like these examples are already not trivially copy-pasteable (since they already use `path`) so the change is not that critical.

A better way to enhance this may be to create something like `cargo-generate` template which would be more customizable and easy to use, nut this surely is out of this PR's scope.

As for moving the examples under `flipperzero` directory -- it may be a good idea, but I would not hurry with this (since it can always be done under a separate task).

JarvisCraft:
> In the past we've talked about whether we should just always enable alloc for the flipperzero crate. From a quick experiment, it looks like just enabling the alloc feature and defining a #[global_allocator] does not have any noticeable effect on the binary size. The main annoyance is that every binary needs to explicitly extern crate flipperzero_alloc, even if not otherwise used.

Yup, currently I have no sane ideas how to enhance this except for a funny macro like:

```rust
#[macro_export]
macro_rules! default_allocator {
    ($crate_name:ident) => {
        extern crate $crate_name;
    };
    () => {
        extern $crate::default_allocator!(flipperzzero_alloc);
    };
}
```

crates/Cargo.toml
JarvisCraft: I think that it's fine to leave it with workspace default as for now and replace in specific crates on demand: for example, most may use workspace's `A.B.C` by using `version.workspace = true` while auto-generated crate would stick to `flipperzero-firmware` version by explicitly specifying it via `version = 'D.E.F-b.G`.

examples/dialog/Cargo.toml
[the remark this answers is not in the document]
dcoles: Using `cargo-generate` was very much what I was thinking (e.g. [`esp-rs` uses this to good effect](https://github.com/esp-rs/esp-template)).

I might take a quick look at how difficult it would be to move the modules. If it turns out to be pretty easy to do, it should cut down the changes required for this PR.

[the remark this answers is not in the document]
JarvisCraft: Amazing! In this case, will be waiting for updates on this :)

JarvisCraft:
I've also bumped the version of the nightly so that newly stabilizzed features (such as sparse registry) can be used.

dcoles:
> I've also bumped the version of the nightly so that newly stabilizzed features (such as sparse registry) can be used.

I was hoping with Rust 1.68.0 we could drop the nightly requirement, but turns out we still need nightly for `-Z no-unique-section-names=yes`. I should ask if it would be possible to stabilize.

crates/flipperzero/examples/gui.rs
```diff
@@ -28,7 +28,7 @@ manifest!(name = "Rust GUI example");
 entry!(main);
 
 /// View draw handler.
-pub extern "C" fn draw_callback(canvas: *mut sys::Canvas, _context: *mut c_void) {
+pub unsafe extern "C" fn draw_callback(canvas: *mut sys::Canvas, _context: *mut c_void) {
```
dcoles, line 31: Was this change intended?

crates/flipperzero/examples/dialog.rs
```diff
@@ -11,8 +11,10 @@ extern crate flipperzero_alloc;
 
 use core::ffi::CStr;
 
-use flipperzero::gui::canvas::Align;
-use flipperzero::dialogs::{self, DialogMessage, DialogMessageButton, DialogsApp};
+use flipperzero::{
```
dcoles, line 14: Was this change intended?

dcoles (approved):

crates/flipperzero/examples/gui.rs
JarvisCraft: Yup, it's due to the fact that while function is `extern "C"`, it still is normally callable from safe thus not having `unsafe` on it (considering the fact that it's dereferencing a raw pointer) is unsound.****

crates/flipperzero/examples/dialog.rs
JarvisCraft: Yes. There was some update of `use` order from Clippy, and then I applied nesting.

dcoles (approved):

dcoles:
Merged. Thanks for the PR @JarvisCraft!

PR #46, opened 2023-02-27
Add a high-level API for interacting with the dolphin

str4d:


crates/flipperzero/src/dolphin/deed.rs
```diff
@@ -0,0 +1,146 @@
+use flipperzero_sys as sys;
+
+/// FlipperZero apps that can generate [`Deed`]s.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum App {
```
dcoles, line 5: Do these enums need to be `#[non_exhaustive]`?

crates/flipperzero/src/dolphin/deed.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn from_raw(raw: sys::DolphinApp) -> Self {
+        match raw {
+            sys::DolphinApp_DolphinAppSubGhz => App::SubGhz,
+            sys::DolphinApp_DolphinAppRfid => App::Rfid,
+            sys::DolphinApp_DolphinAppNfc => App::Nfc,
+            sys::DolphinApp_DolphinAppIr => App::Ir,
+            sys::DolphinApp_DolphinAppIbutton => App::Ibutton,
+            sys::DolphinApp_DolphinAppBadusb => App::Badusb,
+            sys::DolphinApp_DolphinAppPlugin => App::Plugin,
+            _ => panic!("Invalid DolphinApp value {}", raw),
```
dcoles, line 25: It looks like the `DolphinApp` and `DolphinDeed` enums may be extended in the future. Ideally we shouldn't panic in this situation.

str4d: It's unlikely that new apps or deeds would be added without an SDK version bump, but I don't know if it works be a major or minor version bump. IDK if changing enums in C is considered "breaking"; I guess they are effectively non-exhaustive...

Okay, I can do the same trick that I suggested in the filesystem PR. That should let apps automatically handle new apps and deeds that the SDK provides, while preventing them from defining arbitrary app or deed enum values (or the new ones before this crate knows about them).

dcoles (approved):
Thanks for the PR!

PR #47, opened 2023-03-02
feature: Logging

str4d:
The `Level` and `LevelFilter` types are based on the equivalent `tracing-core` types. We then use them to implement wrapper macros around `sys::furi_log_print_format`.


crates/flipperzero/src/furi/log/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+#[macro_export]
+macro_rules! log {
+    (target: $target:expr, $lvl:expr, $msg:expr $(, $arg:expr)*) => ({
+        if $lvl <= $crate::furi::log::LevelFilter::current() {
+            unsafe {
+                $crate::__macro_support::__sys::furi_log_print_format(
+                    $crate::__macro_support::__level_to_furi($lvl),
+                    $crate::__macro_support::__sys::c_string!($target),
+                    $crate::__macro_support::__sys::c_string!($msg),
+                    $($arg,)*
```
str4d, line 31: Currently we just pass in whatever arguments the caller provides. I've confirmed that this does work, but if we spent some time on macro magic we could probably type-check the arguments against the format parameters in `$msg`.

str4d:
Adding `info!("Hello, reader of the logs!");` line to the `hello-rust` example app increases the size of `hello-rust.fap` from 908 bytes to 1080 bytes.

crates/flipperzero/src/furi/log/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+/// log!(target: "events", Level::INFO, "Finished the documentation!");
+/// # }
+/// ```
+#[macro_export]
+macro_rules! log {
+    (target: $target:expr, $lvl:expr, $msg:expr $(, $arg:expr)*) => ({
+        if $lvl <= $crate::furi::log::LevelFilter::current() {
+            unsafe {
+                $crate::__macro_support::__sys::furi_log_print_format(
+                    $crate::__macro_support::__level_to_furi($lvl),
```
str4d, line 28: This forces `$lvl` to have type `Level`.

crates/flipperzero/src/furi/log/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                    $crate::__macro_support::__sys::c_string!($target),
+                    $crate::__macro_support::__sys::c_string!($msg),
```
str4d, line 30: These force `$target` and `$msg` to be literals, though not specifically string literals (instead we get the string representation of whatever literal is passed in, with a trailing null byte).

crates/flipperzero/src/furi/log/metadata.rs
```diff
@@ -0,0 +1,535 @@
...
+                // `set_max` function, which takes a `LevelFilter` parameter;
+                // this restricts the inputs to `set_max` to the set of valid
+                // discriminants. Therefore, **as long as `furi_log_set_level`
+                // is only ever called by `set_max`**, this is safe.
+                core::hint::unreachable_unchecked()
+            },
+        }
+    }
+
+    pub(crate) fn set_max(LevelFilter(level): LevelFilter) {
```
str4d, line 334: Currently this is not exposed; in `tracing-core` it was used to automatically set the application-global maximum level filter based on all `tracing` macro usages in the app and its dependencies. Is it useful to make a public API?

I'll admit to being a bit confused as to why `furi_log_set_level` is in the SDK at all, enabling individual apps to alter the Flipper Zero-global log level rather than requiring it be configured in the settings. I guess so individual apps can temporarily override it and then set it back later?

crates/flipperzero/src/furi/log/metadata.rs
```diff
@@ -0,0 +1,535 @@
...
+                // Flipper Zero SDK to `FuriLogLevelDefault`, which is not a
+                // valid `LevelFilter` discriminant but is specifically handled
+                // above. It is set either internally by the Flipper Zero, or
+                // through the Flipper Zero SDK. The latter we expose via the
+                // `set_max` function, which takes a `LevelFilter` parameter;
+                // this restricts the inputs to `set_max` to the set of valid
+                // discriminants. Therefore, **as long as `furi_log_set_level`
+                // is only ever called by `set_max`**, this is safe.
+                core::hint::unreachable_unchecked()
+            },
```
str4d, line 330: For context, without this `unreachable_unchecked()` (instead relying on `unreachable!()`, and removing its useful error message entirely), `hello-rust.fap` (with its single `info!()` usage) increases from 1080 bytes to 2920 bytes.

examples/hello-rust/src/main.rs
```diff
@@ -24,6 +24,7 @@ entry!(main);
 
 // Entry point
 fn main(_args: *mut u8) -> i32 {
+    info!("Hello, reader of the logs!");
```
dcoles, line 27: Would be nice to show an example of formatting.

crates/flipperzero/src/furi/log/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+/// log!(Level::ERROR, "Failed to handle the florp: %d", error_code);
+/// log!(target: "events", Level::INFO, "Finished the documentation!");
+/// # }
+/// ```
+#[macro_export]
+macro_rules! log {
+    (target: $target:expr, $lvl:expr, $msg:expr $(, $arg:expr)*) => ({
+        if $lvl <= $crate::furi::log::LevelFilter::current() {
+            unsafe {
+                $crate::__macro_support::__sys::furi_log_print_format(
```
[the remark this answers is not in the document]
str4d: Macros by default resolve in the crate they are invoked in; using `::flipperzero_sys` would therefore require downstream users of `flipperzero` to always declare `flipperzero-sys` as a dependency and never rename it.

`$crate` by contrast resolves to the crate the macro is defined in, so it's resilient to both renaming of `flipperzero`, and absence of  `flipperzero-sys` as a direct dependency.

https://doc.rust-lang.org/reference/macros-by-example.html#hygiene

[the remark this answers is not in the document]
dcoles: Thanks for the explanation. I ran into exactly this issue in #48.

dcoles:
@str4d I have a PR up that switches the `print!`/`println!` macros to [`ufmt`](https://crates.io/crates/ufmt). It results in a 4.25x smaller binary for the simple case of `println!("Hello, {}", "Rust")`.

That seems to be enough of an improvement over `core::fmt` that it's probably worth using that rather than calling through to a `printf`-like function for formatting.

crates/flipperzero/src/furi/log/metadata.rs
```diff
@@ -0,0 +1,535 @@
...
+}
+
+impl fmt::Display for Level {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            Level::TRACE => f.pad("TRACE"),
+            Level::DEBUG => f.pad("DEBUG"),
+            Level::INFO => f.pad("INFO"),
+            Level::WARN => f.pad("WARN"),
+            Level::ERROR => f.pad("ERROR"),
```
dcoles, line 146: I think `core::fmt::Formatter::pad` was the function that adds ~3 KiB to the overall binary size.

crates/flipperzero/src/furi/log/metadata.rs
dcoles: Oh nice. I didn't know about `unreachable_unchecked()`. I ended up resorting to a rather hacky solution.

https://github.com/flipperzero-rs/flipperzero/blob/dfb671366e1e19f99e0ed85ac8491fb06bc8bd8a/crates/sys/src/lib.rs#L33

crates/flipperzero/src/furi/log/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+///
+/// # fn main() {
+/// let error_code = 42;
+/// log!(Level::ERROR, "Failed to handle the florp: %d", error_code);
+/// log!(target: "events", Level::INFO, "Finished the documentation!");
+/// # }
+/// ```
+#[macro_export]
+macro_rules! log {
+    (target: $target:expr, $lvl:expr, $msg:expr $(, $arg:expr)*) => ({
```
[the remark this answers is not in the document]
str4d: I was matching the [`log`](https://docs.rs/log/0.4.17/log/macro.log.html) and [`tracing`](https://docs.rs/tracing/0.1.37/tracing/macro.event.html) Rust ecosystems here which align on `target`, with the intent of making it easy to migrate Rust code that already uses one of those. But I'm happy to use `tag` here instead to make it easy to migrate Flipper Zero C code instead.

crates/flipperzero/src/furi/log/mod.rs
str4d: Hmm, a downside to using `ufmt` is that it means logging requires `alloc`, because we need a `String` to accumulate the output into before calling `sys::furi_log_print_format`. Though we could technically avoid enabling `alloc` on the Rust side by using `FuriString` instead? I can implement a Rust wrapper around that.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/fc4eefdabc8eba3537a1ca6ae0413eef1b5daf69..27f9831150924c396e8e45bed25442fa8a31fb30) to fix merge conflicts.

crates/flipperzero/src/furi/log/mod.rs
str4d: Started this in #50.

crates/flipperzero/src/furi/log/mod.rs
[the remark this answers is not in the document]
dcoles: Sounds good. Let's stick with `target` then.

crates/flipperzero/src/furi/log/metadata.rs
```diff
@@ -0,0 +1,535 @@
+//! Metadata describing log data.
+//
+// The structs and enums in this file are extracted from the `tracing-core` crate with
+// adaptions to Furi. The original code is copyright (c) 2019 Tokio Contributors
```
dcoles, line 4: This seems a lot of code for what should mostly be just some macros around `sys::furi_log_print_format`.

The [`log`](https://crates.io/crates/log) crate has `no-std` support and [only needs about 15 lines to implement](https://docs.rs/log/0.4.17/log/#implementing-a-logger). As a bonus, we'd immediately support any crate that used the `log` crate as a logging facade.

The [`tracing`](https://crates.io/crates/tracing) crate is quite powerful, but the main reason for using it is to be able to understand asynchronous code where tasks can jump between threads. While it looks like it could be possible to implement a [`tracing::Subscriber`](https://docs.rs/tracing/0.1.37/tracing/trait.Subscriber.html), there's a lot of code that needs to be generated to support spans, which I worry would blow out binary sizes quite quickly.

crates/flipperzero/src/furi/log/metadata.rs
dcoles: I'm not 100% sure, but I think the current Flipper Zero firmware is built with the Memory Protection Unit (MPU) disabled, thus all threads exist in a single memory space. So applications are really just a thread in the flat memory space.

crates/flipperzero/src/furi/log/metadata.rs
str4d: I actually started this effort by trying to write a `log` backend, and I very quickly ran into the problem that `log` assumes that you will be writing and formatting with `&str`, while `sys::furi_log_print_format` requires `&CStr`. This would force inline allocations at every individual log point, in order to copy the format string and append a `nul` byte.

`tracing` has the same issue here. It's a fantastic library and I use it wherever I can (it's useful for way more than just understanding async code), but it also is built around `&str`, and thus doesn't match up well with `sys::furi_log_print_format`.

My next thought was that we can use the `c_string!` macro to ensure that the format string is statically null-terminated, but this requires writing our own macros. Once we've done that, the next step is to represent the `FuriLevel` enum on the Rust side, which is where I turned to `tracing`'s implementation as it is heavily optimised for low code overhead, which we want here.

crates/flipperzero/src/furi/log/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        if $lvl <= $crate::furi::log::LevelFilter::current() {
```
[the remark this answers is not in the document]
str4d: That is indeed currently the case, but that assumes we continue to rely on the C side for `printf` formatting. Once we use `ufmt` and `FuriString` to format the log output before calling `sys::furi_log_print_format`, we will want to avoid performing any of that if we don't have to.

crates/flipperzero/src/furi/log/metadata.rs
str4d: All that being said: if we're going to use `ufmt` and `FuriString` for formatting log output before calling `sys::furi_log_print_format`, then we _could_ provide a `log::Log` and/or `tracing::Subscriber` implementation that users can choose to install.

dcoles: I think `FuriString` is going to be the right approach. We _may_ be able to do something clever like using a static `FuriString` buffer for formatting, so we're not constantly doing allocations (the logger is already [serialized on a mutex](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@61744b8d01e7ad1d7544a6a0176fac5369e67dc4/-/blob/furi/core/log.c?L27)).

I just realized another issue with the `log` crate. It expects [records to be formatted as `core::fmt::Arguments`](https://docs.rs/log/0.4.17/log/struct.Record.html#method.args) which pretty much forces the use of `core::fmt` which we're trying to avoid. Tracing is a bit better since it can keep track of individual fields, but generic messages are still formatted using `core::fmt`.

str4d:
[Rebased on current `main`](https://github.com/flipperzero-rs/flipperzero/compare/27f9831150924c396e8e45bed25442fa8a31fb30..cd0d07bc7548a968190c0f6ec63107638f8d3a56). I am not currently working on this (because it is blocked on #50, which in turn is blocked on #57), but I will rebase it for testing purposes when the `main` branch is updated to newer SDKs.

str4d:
[Rebased on `aa1bfde5f3ce688c1f44c6945ea5863d03b5dac0`](https://github.com/flipperzero-rs/flipperzero/compare/cd0d07bc7548a968190c0f6ec63107638f8d3a56..00e1ae8fed8f24b20616a1136b2dc630238ab193) (the same base as #50).

crates/flipperzero/src/furi/log/mod.rs
str4d: I have this working locally.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/00e1ae8fed8f24b20616a1136b2dc630238ab193..d9f8b3cdec92f7798383b53af1b5bcf555f76a00) and switched from the Flipper Zero SDK's `printf` to `FuriString` and `ufmt` instead.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/d9f8b3cdec92f7798383b53af1b5bcf555f76a00..5af2c2788a380d184cb864cb36ad39ca4e4cde53) because I forgot I'd renamed `String` to `FuriString` 😅

crates/flipperzero/src/furi/log/metadata.rs
str4d: I've added implementations of `ufmt::{uDebug, uDisplay}` that can be used instead.

crates/flipperzero/src/furi/log/mod.rs
str4d: We now use `ufmt::uwrite!` to constrain `$msg` instead.

crates/flipperzero/src/furi/log/metadata.rs
str4d: I decided to remove this method. If we want to expose `sys::furi_log_set_level`, it should probably be as a standalone method instead to more clearly delineate FAP logging from Flipper Zero-global logging.

crates/flipperzero/src/furi/log/metadata.rs
str4d: We now use `FuriString` (allocating each time), and `ufmt` for formatting. So our arguments are whatever `ufmt::uwrite` takes instead of `core::fmt::Arguments`.

We could maybe use the `rt` macros to set up a static, and then free it on exit. That seems like an approach that would be generally useful for various sub-systems; opened #64.

PR #49, opened 2023-03-03
Implement `std::thread`-like wrappers around Furi Thread APIs

str4d:


crates/flipperzero/examples/threads.rs
```diff
@@ -0,0 +1,65 @@
...
+        .name("Flipper".to_owned())
+        .expect("name is valid")
+        .spawn(|| {
+            println!("Second thread started!");
+            thread::sleep(Duration::from_secs(2));
+            println!("Second thread finished!");
+            0
+        });
+
+    for (i, thread) in [&thread::current(), first.thread(), second.thread()]
```
str4d, line 49: Currently, if `thread::current()` is included in this list, I reliably crash Flipper Zero with a Null Pointer Dereference. I suspect it is because I'm freeing the `*mut sys::FuriThread` returned by `sys::furi_thread_get_current()`, which implies I don't currently fully understand how `FuriThread` memory is meant to be managed.

dcoles: That's interesting. I'm surprised it doesn't crash first from the `furi_assert(thread->state == FuriThreadStateStopped);`.

I think you'll need a boolean to keep track of whether you own this thread or it's a foreign one.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -13,3 +136,132 @@ pub fn sleep(duration: core::time::Duration) {
         }
     }
 }
+
+/// A unique identifier for a running thread.
+#[cfg(feature = "alloc")]
+pub struct ThreadId(sys::FuriThreadId);
```
dcoles, line 142: `sys::FuriThreadId` is really an alias for FreeRTOS's `TaskHandle_t`. There might be some time that you need this for talking with the low-level FreeRTOS APIs, so you probably want to use `#[repr(transparent)]` or just make `ThreadId` a type alias.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -13,3 +136,132 @@ pub fn sleep(duration: core::time::Duration) {
...
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl ufmt::uDebug for Thread {
+    fn fmt<W>(&self, f: &mut ufmt::Formatter<'_, W>) -> Result<(), W::Error>
+    where
+        W: ufmt::uWrite + ?Sized,
+    {
+        // TODO: ufmt doesn't provide an impl of uDebug for &str.
```
dcoles, line 218: That seems quite the omission. 😦

crates/flipperzero/src/furi/thread.rs
```diff
@@ -13,3 +136,132 @@ pub fn sleep(duration: core::time::Duration) {
...
+pub struct Thread {
+    /// Guaranteed to be UTF-8.
+    name: Option<CString>,
+    thread: *mut sys::FuriThread,
+}
+
+#[cfg(feature = "alloc")]
+impl Drop for Thread {
+    fn drop(&mut self) {
+        // TODO: is this safe to do while the thread is running?
```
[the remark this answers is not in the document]
str4d: Right; I wasn't sure what managed the memory of the thread handles, but this combined with the crash from freeing the result of `sys::furi_thread_get_current()` suggests that the handles are all just pointers to the same memory. So at a minimum we need to wrap it in an `Arc` internally on the Rust side.

We also need to handle the fact that the SDK appears to expect us to clean up the thread memory that we allocate ourselves, but needs us to not do so until the firmware has finished using it.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -13,3 +136,132 @@ pub fn sleep(duration: core::time::Duration) {
...
+
+/// A unique identifier for a running thread.
+#[cfg(feature = "alloc")]
+pub struct ThreadId(sys::FuriThreadId);
+
+/// A handle to a thread.
+#[cfg(feature = "alloc")]
+pub struct Thread {
+    /// Guaranteed to be UTF-8.
+    name: Option<CString>,
```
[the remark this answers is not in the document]
str4d: I originally considered doing this, and then switched to the same approach that `std::thread::Thread` uses internally, so that the names were guaranteed to be UTF-8 (thus enabling us to have `Thread::name() -> Option<&str>` cheaply). This will always be the case for names of threads created in Rust, but isn't guaranteed to be the case for `sys::furi_thread_get_current()`.

We could instead use the name stored in the `FuriThread`, but we'd need to do one of:
- Return `&CStr` instead of `&str`.
- Check on every call that it is valid UTF-8, even if we know that it is because we created it.
- Cache some `bool` inside the `Thread` object, so we only pay the cost of checking for valid UTF-8 on first usage, and then leverage the faster path on subsequent calls.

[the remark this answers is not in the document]
dcoles: Ah. OK. That makes sense why `std::thread::Thread` stores its own copy. It can't trust that the OS hasn't tainted it with non-UTF-8 values, even if it was the one who set it when the thread was created.

OK. I understand the current behaviour now. Thanks.

crates/flipperzero/src/furi/thread.rs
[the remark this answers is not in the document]
dcoles: That's what it looks like. `sys::furi_thread_get_current()` just pulls a copy of the original `FuriThread*` that gets stashed in Thread Local Storage when the thread starts. Even the [main thread is initialised in this way](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@5d4057f7228b528896578ab6cf9d9a61b80a7c58/-/blob/firmware/targets/f7/src/main.c?L29:5).

This present quite a challenge, since you can't call `sys::furi_thread_free` until the thread is stopped, and the only way to guarantee that is by calling `sys::furi_thread_join`.

At present, the only way I can see of doing this safely and without potentially leaking memory is by using some sort of [scoped threads](https://doc.rust-lang.org/std/thread/fn.scope.html) that ensures that all threads spawned under the scope are joined (and `free`'d) before exiting the scope.

Since everything is in the same flat linear memory space, Furi can't even stop/cleanup all the threads after our app "exits". Pretty much the only option is to call `sys::furi_crash`.

crates/flipperzero/src/furi/thread.rs
str4d: The upstream issue is https://github.com/japaric/ufmt/issues/52.

str4d:
[Rebased on current `main`](https://github.com/flipperzero-rs/flipperzero/compare/858fd2064447e339ae3d8706c11ac1cc7c4fb782..42eab806f5be230432b909d0c237c7a53c06e102).

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/42eab806f5be230432b909d0c237c7a53c06e102..8b8da1c0a02ff0a6e1c976c712bac3caf87fdc62) to fix thread memory freeing and add `Builder::enable_heap_trace`.

crates/flipperzero/examples/threads.rs
str4d: I'm now doing this by never calling `sys::furi_thread_free` from `Thread` directly, but only inside either `impl Drop for JoinHandle` or in the thread's `FuriThreadStateStopped` handler.

crates/flipperzero/src/furi/thread.rs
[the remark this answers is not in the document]
str4d: I've resolved the primary issue with the introduction of `FreeMan`, so we now never attempt to free a `*mut sys::FuriThread` we are holding while the thread is running.

crates/flipperzero/examples/threads.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    // We can either drop the `JoinHandle` and let the thread complete in the background,
+    // or we can join the thread which blocks until its completion.
+    drop(second);
+    first.join();
+
+    // If we dropped the `JoinHandle` for `first` instead of `second` (or if we dropped
+    // both) we get a null pointer dereference, because it outlives the main thread.
+    // drop(first);
+    // second.join();
```
str4d, line 68: This is the remaining issue that `FreeMan` doesn't currently solve: we either need some way for the main thread to implicitly join on any unfinished threads, or we need those threads to crash "safely" without producing a NPD.

str4d: Interesting, I was also able to get a `BusFault` error when triggering this. In any case, adding a `thread::sleep(Duration::from_secs(4))` after these two lines fixes the crash as expected (because the main thread waits for a total of 7 seconds, while `first` runs for 5).

dcoles: So as far as I can tell, there is no safe way to exit a thread other than returning from main (`vTaskDelete` is not exposed and any abnormal exit calls `furi_thread_catch` which triggers a "You are doing it wrong" crash).

We could probably use `furi_thread_set_state_callback` to register a `FuriThreadStateStopped` hook for the main thread that ensures it doesn't exit until there are no other threads with the same `appid` running. Since the [loader also uses the same state callback](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@e42aec68c533e6120f0e776e8bdc5d7d89e99c8b/-/blob/applications/services/loader/loader.c?L300), we'll probably need to call that afterwards.

You could possibly argue that this is something that the loader should do, since it's the one about to pull the rug out from all these threads.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -1,7 +1,194 @@
 //! Furi Thread API.
 
+#[cfg(feature = "alloc")]
```
dcoles, line 3: Since this can't really be used without `alloc`, we could move the `#[cfg(feature = "alloc")]` to the module definition in `furi/mod.rs`.

str4d: A few standalone functions can be used without `alloc` (e.g. `thread::sleep`).

dcoles: Ah. Sorry. I was sleepy last night.

crates/flipperzero/examples/threads.rs
str4d: > We could probably use `furi_thread_set_state_callback` to register a `FuriThreadStateStopped` hook for the main thread that ensures it doesn't exit until there are no other threads with the same `appid` running. Since the [loader also uses the same state callback](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@e42aec68c533e6120f0e776e8bdc5d7d89e99c8b/-/blob/applications/services/loader/loader.c?L300), we'll probably need to call that afterwards.

The state callback [can only be set while the thread is stopped](https://github.com/flipperdevices/flipperzero-firmware/blob/238005890ec6a619157ac75eb60bd02dfe6c736e/furi/core/thread.c#L229-L233), and there is no API to extract the current handler.

dcoles: Sounds good!

Might also be a good idea to print a warning in the logs, just in case people wonder why their app is hanging at exit.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -1,7 +1,194 @@
...
+        }
+        let callback: sys::FuriThreadCallback = Some(run_thread_body);
+        let context = Box::into_raw(thread_body);
+
+        let freeman = FreeMan::new();
+
+        struct StateContext {
+            thread: *mut sys::FuriThread,
+            freeman: Arc<Mutex<FreeMan>>,
+        }
```
dcoles, line 101: I was thinking about this and was wondering if we could achieve the desired behaviour with just `Arc`.

In `Thread::spawn`:
```rust
pub fn spawn<F>(self, f: F) -> JoinHandle
where
        F: FnOnce() -> i32,
        F: Send + 'static,
{
    ...

    let context = Arc::new(ThreadContext {
        thread: *mut sys::FuriThread,
        // anything else you want to remember (e.g. the UTF-8 thread name)
    });
    let join_handle = JoinHandle {
        context: Arc::clone(&thread_context);
    };
    unsafe {
        sys::furi_thread_set_state_context(thread.thread, Arc::into_raw(context) as *mut c_void);
    }
    
    ...
    
    join_handle
}
```

In `run_state_callback`:
```rust
unsafe extern "C" fn run_state_callback(state: sys::FuriThreadState, context: *mut c_void) {
    if state == sys::FuriThreadState_FuriThreadStateStopped {
        let context = unsafe { Arc::from_raw(context as *mut ThreadContext) };
        if let Some(context) = context.into_inner() {
            // We were able to successfully extract the thread context from `Arc`, thus there is no other refs
            unsafe { sys::furi_thread_free(context.thread) };
        }
    }
}
```

In `ThreadHandle::join`:
```rust
pub fn join(self) -> i32 {
    unsafe {
        sys::furi_thread_join(self.context.thread);

        sys::furi_thread_get_return_code(self.context.thread)
    }
}
```

In `ThreadHandle::drop`:
```rust
fn drop(&mut self) {
    if let Some(context) = context.into_inner() {
        // We were able to successfully extract the thread context from `Arc`, thus there is no other refs
        unsafe { sys::furi_thread_free(context.thread) };
    }
}
```

By using `Arc` as a reference counter, we can avoid the need for keeping track of whether the thread has been stopped or dropped.

I think we could even potentially `ThreadContext` for `Thread::current`, but you would have to call [`Arc::increment_strong_count`](https://doc.rust-lang.org/alloc/sync/struct.Arc.html#method.increment_strong_count) after `Arc::from_raw` to ensure we didn't accidentally free the context when that `Arc`-from-context-pointer dropped.

str4d: Oh, I didn't notices `Arc::from_inner`, that does work quite nicely. I also originally thought the logic would be more complex than it ended up being.

str4d: Oh wait, `Arc::into_inner` is an experimental nightly feature, so that would increase the number of things we are blocked on for #58. And `Arc::strong_count` is insufficient because it doesn't lock, so we would need a `Mutex` around the `Arc`, and then we are getting sufficiently complex that the `FreeMan` solution looks fine.

crates/flipperzero/examples/threads.rs
str4d: Done.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/841105f5b21fe7d89b0e06e7912e5496f8f361f4..0d27779cbc87066eda94ec4a9038d847739fb7ef) to add logging around the thread waiter logic.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/7156b33c266c42bbaf1797936f73e5ce42e6877b..e74cac2df3b0b09984ca87d842e1eed078f4a940) to bump the pinned nightly to one where `Arc::into_inner` is stabilized.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/e74cac2df3b0b09984ca87d842e1eed078f4a940..4211c9b82eb11ed7d27eb0e6e79fb5513215ac1c) to use `sys::c_string!()` instead of manually crafting the log messages for the thread waiter.

PR #50, opened 2023-03-04
Implement `String`-like wrapper around `FuriString`

str4d:
`std::string::String` methods to be implemented:
- [x] `String::new() -> String`
- [x] `String::with_capacity(capacity: usize) -> String`
- [ ] `String::from_utf8(vec: Vec<u8>) -> Result<String, FromUtf8Error>`
- [ ] `String::from_utf8_lossy(v: &[u8]) -> Cow<'_, str>`
- [ ] `String::from_utf16(v: &[u16]) -> Result<String, FromUtf16Error>`
- [ ] `String::from_utf16_lossy(v: &[u16]) -> String`
- ~~`String::into_raw_parts(self) -> (*mut u8, usize, usize)`~~
  - Doesn't make sense for `FuriString`. We could maybe expose `*mut sys::FuriString`?
- ~~`String::from_raw_parts(buf: *mut u8, length: usize, capacity: usize) -> String`~~
  - We'd only expose this (via `unsafe`) if we want to leave `*mut sys::FuriString` exposed.
- ~~`String::from_utf8_unchecked(bytes: Vec<u8>) -> String`~~
  - Doesn't make sense for `FuriString`.
- ~~`String::into_bytes(self) -> Vec<u8>`~~
  - Doesn't make sense for `FuriString` (we aren't storing the string internally as a `Vec<u8>`, so it's better to just call `String::to_bytes().to_owned()`).
- ~~`String::as_str(&self) -> &str`~~
  - [x] `String::as_c_str(&self) -> &CStr`
- ~~`String::as_mut_str(&mut self) -> &mut str`~~
- [x] `String::push_str(&mut self, string: &str)`
- ~~`String::capacity(&self) -> usize`~~
  - Flipper Zero SDK doesn't expose the necessary method.
- [x] `String::reserve(&mut self, additional: usize)`
- ~~`String::reserve_exact(&mut self, additional: usize)`~~
  - Flipper Zero SDK doesn't expose the necessary method.
- ~~`String::try_reserve(&mut self, additional: usize) -> Result<(), TryReserveError>`~~
  - Flipper Zero SDK doesn't expose fallible `FuriString` allocations.
- ~~`String::try_reserve_exact(&mut self, additional: usize) -> Result<(), TryReserveError>`~~
  - Flipper Zero SDK doesn't expose fallible `FuriString` allocations.
- ~~`String::shrink_to_fit(&mut self)`~~
  - Flipper Zero SDK doesn't expose the necessary method.
- ~~`String::shrink_to(&mut self, min_capacity: usize)`~~
  - Flipper Zero SDK doesn't expose the necessary method.
- [x] `String::push(&mut self, ch: char)`
- [x] `String::as_bytes(&self) -> &[u8]`
  - Renamed to `String::to_bytes` because it's a bit more expensive.
  - Also added `String::to_bytes_with_nul` for parity with `CString`.
- [x] `String::truncate(&mut self, new_len: usize)`
- [ ] `String::pop(&mut self) -> Option<char>`
- [ ] `String::remove(&mut self, idx: usize) -> char`
- [ ] `String::retain<F>(&mut self, mut f: F) where F: FnMut(char) -> bool`
- [x] `String::insert(&mut self, idx: usize, ch: char)`
- [x] `String::insert_str(&mut self, idx: usize, string: &str)`
- ~~`String::as_mut_vec(&mut self) -> &mut Vec<u8>`~~
  - Doesn't make sense for `FuriString`.
- [x] `String::len(&self) -> usize`
- [x] `String::is_empty(&self) -> bool`
- [x] `String::split_off(&mut self, at: usize) -> String`
- [x] `String::clear(&mut self)`
- [ ] `String::drain<R>(&mut self, range: R) -> Drain<'_> where R: RangeBounds<usize>`
- [ ] `String::replace_range<R>(&mut self, range: R, replace_with: &str) where R: RangeBounds<usize>`
- ~~`String::into_boxed_str(self) -> Box<str>`~~
  - Doesn't make sense for `FuriString`.
- [x] `impl Clone for String`
- [x] `impl FromIterator<char> for String`
- [x] `impl<'a> FromIterator<&'a char> for String`
- [x] `impl<'a> FromIterator<&'a str> for String`
- [x] `impl FromIterator<String> for String`
- [x] `impl FromIterator<Box<str>> for String`
- [x] `impl<'a> FromIterator<Cow<'a, str>> for String`
- [x] `impl Extend<char> for String`
- [x] `impl<'a> Extend<&'a char> for String`
- [x] `impl<'a> Extend<&'a str> for String`
- [x] `impl Extend<Box<str>> for String`
- [x] `impl Extend<String> for String`
- [x] `impl<'a> Extend<Cow<'a, str>> for String`
- [x] `impl PartialEq for String`
- [x] `impl Eq for String`
- [x] `impl PartialOrd for String`
- [x] `impl Ord for String`
- [x] `impl Default for String`
- [x] `impl fmt::Display for String`
- [x] `impl fmt::Debug for String`
- [x] `impl hash::Hash for String`
- [x] `impl Add<&str> for String`
- [x] `impl AddAssign<&str> for String`
- [ ] `impl ops::Index<ops::Range<usize>> for String`
- [ ] `impl ops::Index<ops::RangeTo<usize>> for String`
- [ ] `impl ops::Index<ops::RangeFrom<usize>> for String`
- [ ] `impl ops::Index<ops::RangeFull> for String`
- [ ] `impl ops::Index<ops::RangeInclusive<usize>> for String`
- [ ] `impl ops::Index<ops::RangeToInclusive<usize>> for String`
- [ ] `impl ops::IndexMut<ops::Range<usize>> for String`
- [ ] `impl ops::IndexMut<ops::RangeTo<usize>> for String`
- [ ] `impl ops::IndexMut<ops::RangeFrom<usize>> for String`
- [ ] `impl ops::IndexMut<ops::RangeFull> for String`
- [ ] `impl ops::IndexMut<ops::RangeInclusive<usize>> for String`
- [ ] `impl ops::IndexMut<ops::RangeToInclusive<usize>> for String`
- [ ] `impl ops::Deref for String`
- [ ] `impl ops::DerefMut for String`
- [ ] `impl FromStr for String`
- ~~`impl AsRef<str> for String`~~
  - [x] `impl AsRef<CStr> for String`
- ~~`impl AsMut<str> for String`~~
- [ ] `impl AsRef<[u8]> for String`
- [x] `impl From<&str> for String`
- [x] `impl From<&mut str> for String`
- [ ] `impl From<&String> for String`
- [x] `impl From<Box<str>> for String`
- ~~`impl From<String> for Box<str>`~~
  - Doesn't make sense for `FuriString`.
- [x] `impl<'a> From<Cow<'a, str>> for String`
- ~~`impl<'a> From<String> for Cow<'a, str>`~~
  - Doesn't make sense for `FuriString`.
- ~~`impl<'a> From<&'a String> for Cow<'a, str>`~~
  - Doesn't make sense for `FuriString`.
- ~~`impl<'a> FromIterator<String> for Cow<'a, str>`~~
  - Doesn't make sense for `FuriString`.
- ~~`impl From<String> for Vec<u8>`~~
  - Doesn't make sense for `FuriString`.
- [x] `impl fmt::Write for String`
- [x] `impl From<char> for String`

`str` methods to be implemented:
- ~~`str::len(&self) -> usize`~~ (see `String`)
- ~~`str::is_empty(&self) -> bool`~~ (see `String`)
- [ ] `str::is_char_boundary(&self, index: usize) -> bool`
- [ ] `str::floor_char_boundary(&self, index: usize) -> usize`
- [ ] `str::ceil_char_boundary(&self, index: usize) -> usize`
- ~~`str::as_bytes(&self) -> &[u8]`~~ (see `String`)
- [ ] `str::as_bytes_mut(&mut self) -> &mut [u8]`
- [ ] `str::as_ptr(&self) -> *const u8`
- [ ] `str::as_mut_ptr(&mut self) -> *mut u8`
- [ ] `str::get<I: ~const SliceIndex<str>>(&self, i: I) -> Option<&I::Output>`
- [ ] `str::get_mut<I: ~const SliceIndex<str>>(&mut self, i: I) -> Option<&mut I::Output>`
- [ ] `str::get_unchecked<I: ~const SliceIndex<str>>(&self, i: I) -> &I::Output`
- [ ] `str::get_unchecked_mut<I: ~const SliceIndex<str>>(&mut self, i: I) -> &mut I::Output`
- ~~`str::slice_unchecked(&self, begin: usize, end: usize) -> &str`~~ (deprecated)
- ~~`str::slice_mut_unchecked(&mut self, begin: usize, end: usize) -> &mut str`~~ (deprecated)
- [ ] `str::split_at(&self, mid: usize) -> (&str, &str)`
- [ ] `str::split_at_mut(&mut self, mid: usize) -> (&mut str, &mut str)`
- [x] `str::chars(&self) -> Chars<'_>`
  - Renamed to `FuriString::chars_lossy` because it inserts replacement characters.
- [x] `str::char_indices(&self) -> CharIndices<'_>`
  - Renamed to `FuriString::chars_lossy` because it inserts replacement characters.
- [x] `str::bytes(&self) -> Bytes<'_>`
- [ ] `str::split_whitespace(&self) -> SplitWhitespace<'_>`
- [ ] `str::split_ascii_whitespace(&self) -> SplitAsciiWhitespace<'_>`
- [ ] `str::lines(&self) -> Lines<'_>`
- ~~`str::lines_any(&self) -> LinesAny<'_>`~~ (deprecated)
- [ ] `str::encode_utf16(&self) -> EncodeUtf16<'_>`
- [x] `str::contains<'a, P: Pattern<'a>>(&'a self, pat: P) -> bool`
- [x] `str::starts_with<'a, P: Pattern<'a>>(&'a self, pat: P) -> bool`
- [x] `str::ends_with<'a, P>(&'a self, pat: P) -> bool`
- [x] `str::find<'a, P: Pattern<'a>>(&'a self, pat: P) -> Option<usize>`
- [x] `str::rfind<'a, P>(&'a self, pat: P) -> Option<usize>`
- [ ] `str::split<'a, P: Pattern<'a>>(&'a self, pat: P) -> Split<'a, P>`
- [ ] `str::split_inclusive<'a, P: Pattern<'a>>(&'a self, pat: P) -> SplitInclusive<'a, P>`
- [ ] `str::rsplit<'a, P>(&'a self, pat: P) -> RSplit<'a, P>`
- [ ] `str::split_terminator<'a, P: Pattern<'a>>(&'a self, pat: P) -> SplitTerminator<'a, P>`
- [ ] `str::rsplit_terminator<'a, P>(&'a self, pat: P) -> RSplitTerminator<'a, P>`
- [ ] `str::splitn<'a, P: Pattern<'a>>(&'a self, n: usize, pat: P) -> SplitN<'a, P>`
- [ ] `str::rsplitn<'a, P>(&'a self, n: usize, pat: P) -> RSplitN<'a, P>`
- [ ] `str::split_once<'a, P: Pattern<'a>>(&'a self, delimiter: P) -> Option<(&'a str, &'a str)>`
- [ ] `str::rsplit_once<'a, P>(&'a self, delimiter: P) -> Option<(&'a str, &'a str)>`
- [ ] `str::matches<'a, P: Pattern<'a>>(&'a self, pat: P) -> Matches<'a, P>`
- [ ] `str::rmatches<'a, P>(&'a self, pat: P) -> RMatches<'a, P>`
- [ ] `str::match_indices<'a, P: Pattern<'a>>(&'a self, pat: P) -> MatchIndices<'a, P>`
- [ ] `str::rmatch_indices<'a, P>(&'a self, pat: P) -> RMatchIndices<'a, P>`
- [x] `str::trim(&self) -> &str`
  - Made mutable.
- [x] `str::trim_start(&self) -> &str`
  - Made mutable.
- [x] `str::trim_end(&self) -> &str`
  - Made mutable.
- ~~`str::trim_left(&self) -> &str`~~ (deprecated)
- ~~`str::trim_right(&self) -> &str`~~ (deprecated)
- [x] `str::trim_matches<'a, P>(&'a self, pat: P) -> &'a str`
  - Made mutable.
- [x] `str::trim_start_matches<'a, P: Pattern<'a>>(&'a self, pat: P) -> &'a str`
  - Made mutable.
- [x] `str::strip_prefix<'a, P: Pattern<'a>>(&'a self, prefix: P) -> Option<&'a str>`
  - Made mutable and returns `bool`.
- [x] `str::strip_suffix<'a, P: Pattern<'a>>(&'a self, suffix: P) -> Option<&'a str>`
  - Made mutable and returns `bool`.
- [x] `str::trim_end_matches<'a, P>(&'a self, pat: P) -> &'a str`
  - Made mutable.
- ~~`str::trim_left_matches<'a, P: Pattern<'a>>(&'a self, pat: P) -> &'a str`~~ (deprecated)
- ~~`str::trim_right_matches<'a, P>(&'a self, pat: P) -> &'a str`~~ (deprecated)
- [ ] `str::parse<F: FromStr>(&self) -> Result<F, F::Err>`
- [ ] `str::is_ascii(&self) -> bool`
- [ ] `str::eq_ignore_ascii_case(&self, other: &str) -> bool`
- [ ] `str::make_ascii_uppercase(&mut self)`
- [ ] `str::make_ascii_lowercase(&mut self)`
- [ ] `str::escape_debug(&self) -> EscapeDebug<'_>`
- [ ] `str::escape_default(&self) -> EscapeDefault<'_>`
- [ ] `str::escape_unicode(&self) -> EscapeUnicode<'_>`

### DIFF
--- a/crates/flipperzero/src/furi/string.rs
+++ b/crates/flipperzero/src/furi/string.rs
@@ -1,0 +1,575 @@
+//! String primitives built around `FuriString`.
+
+use core::{
+    cmp::Ordering,
+    convert::Infallible,
+    ffi::{c_char, CStr},
+    fmt::{self, Write},
+    hash,
+    ops::{Add, AddAssign},
+    ptr,
+};
+
+#[cfg(feature = "alloc")]
+use alloc::{borrow::Cow, boxed::Box, ffi::CString};
+
+use flipperzero_sys as sys;
+
+/// A Furi string.
+///
+/// This is similar to Rust's [`CString`] in that it represents an owned, C-compatible,
+/// nul-terminated string with no nul bytes in the middle. It also has additional methods
+/// to provide the flexibility of Rust's [`String`]. It is used in various APIs of the
+/// Flipper Zero SDK.
+///
+/// This type does not requre the `alloc` feature flag, because it does not use the Rust
+/// allocator. Very short strings (7 bytes or fewer) are stored directly inside the
+/// `FuriString` struct (which is stored on the heap), while longer strings are allocated
+/// on the heap by the Flipper Zero firmware.
+#[derive(Eq)]
+pub struct String(*mut sys::FuriString);
+
+impl Drop for String {
+    fn drop(&mut self) {
+        unsafe { sys::furi_string_free(self.0) };
+    }
+}
+
+// Implementations matching `std::string::String`.
+impl String {
+    /// Creates a new empty `String`.
+    #[inline]
+    #[must_use]
+    pub fn new() -> Self {
+        String(unsafe { sys::furi_string_alloc() })
+    }
+
+    /// Creates a new empty `String` with at least the specified capacity.
+    #[inline]
+    #[must_use]
+    pub fn with_capacity(capacity: usize) -> Self {
+        let s = Self::new();
+        // The size passed to `sys::furi_string_reserve` needs to include the nul
+        // terminator.
+        unsafe { sys::furi_string_reserve(s.0, capacity + 1) };
+        s
+    }
+
+    #[inline]
+    #[must_use]
+    fn as_c_ptr(&self) -> *const c_char {
+        unsafe { sys::furi_string_get_cstr(self.0) }
+    }
+
+    /// Extracts a `CStr` containing the entire string slice, with nul termination.
+    #[inline]
+    #[must_use]
+    pub fn as_c_str(&self) -> &CStr {
+        unsafe { CStr::from_ptr(self.as_c_ptr()) }
+    }
+
+    /// Appends a given `String` onto the end of this `String`.
+    #[inline]
+    pub fn push_string(&mut self, string: &String) {
+        unsafe { sys::furi_string_cat(self.0, string.0) }
+    }
+
+    /// Appends a given `str` onto the end of this `String`.
+    #[inline]
+    pub fn push_str(&mut self, string: &str) {
+        self.extend(string.chars());
+    }
+
+    /// Appends a given `CStr` onto the end of this `String`.
+    #[inline]
+    pub fn push_c_str(&mut self, string: &CStr) {
+        unsafe { sys::furi_string_cat_str(self.0, string.as_ptr()) }
+    }
+
+    /// Reserves capacity for at least `additional` bytes more than the current length.
+    #[inline]
+    pub fn reserve(&mut self, additional: usize) {
+        // `self.len()` counts the number of bytes excluding the terminating nul, but the
+        // size passed to `sys::furi_string_reserve` needs to include the nul terminator.
+        unsafe { sys::furi_string_reserve(self.0, self.len() + additional + 1) };
+    }
+
+    /// Appends the given [`char`] to the end of this `String`.
+    #[inline]
+    pub fn push(&mut self, ch: char) {
+        match ch.len_utf8() {
+            1 => unsafe { sys::furi_string_push_back(self.0, ch as c_char) },
+            _ => unsafe { sys::furi_string_utf8_push(self.0, ch as u32) },
+        }
+    }
+
+    /// Returns a byte slice of this `String`'s contents.
+    ///
+    /// The returned slice will **not** contain the trailing nul terminator that the
+    /// underlying C string has.
+    #[inline]
+    #[must_use]
+    pub fn to_bytes(&self) -> &[u8] {
+        self.as_c_str().to_bytes()
+    }
+
+    /// Returns a byte slice of this `String`'s contents with the trailing nul byte.
+    ///
+    /// This function is the equivalent of [`String::to_bytes`] except that it will retain
+    /// the trailing nul terminator instead of chopping it off.
+    #[inline]
+    #[must_use]
+    pub fn to_bytes_with_nul(&self) -> &[u8] {
+        self.as_c_str().to_bytes_with_nul()
+    }
+
+    /// Shortens this `String` to the specified length.
+    ///
+    /// If `new_len` is greater than the string's current length, this has no effect.
+    #[inline]
+    pub fn truncate(&mut self, new_len: usize) {
+        unsafe { sys::furi_string_left(self.0, new_len) };
+    }
+
+    /// Inserts a character into this `String` at a byte position.
+    ///
+    /// This is an *O*(*n*) operation as it requires copying every element in the buffer.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `idx` is larger than the `String`'s length.
+    #[inline]
+    pub fn insert(&mut self, idx: usize, ch: char) {
+        let mut buf = [0; 4];
+        let encoded = ch.encode_utf8(&mut buf).as_bytes();
+        self.insert_bytes(idx, encoded);
+    }
+
+    /// Inserts a string slice into this `String` at a byte position.
+    ///
+    /// This is an *O*(*n*) operation as it requires copying every element in the buffer.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `idx` is larger than the `String`'s length.
+    #[inline]
+    pub fn insert_str(&mut self, idx: usize, string: &str) {
+        self.insert_bytes(idx, string.as_bytes());
+    }
+
+    fn insert_bytes(&mut self, idx: usize, bytes: &[u8]) {
+        let len = self.len();
+        assert!(idx <= len);
+
+        // Reserve sufficient space to insert `bytes` without repeat re-allocations.
+        let amt = bytes.len();
+        self.reserve(amt);
+
+        // Append `bytes` to force the underlying `FuriString` to be the correct length.
+        for byte in bytes.iter() {
+            unsafe { sys::furi_string_push_back(self.0, *byte as c_char) };
+        }
+
+        // Now use pointer access to construct the expected `String` contents.
+        let c_ptr = self.as_c_ptr().cast::<u8>();
+        unsafe {
+            ptr::copy(c_ptr.add(idx), c_ptr.cast_mut().add(idx + amt), len - idx);
+            ptr::copy_nonoverlapping(bytes.as_ptr(), c_ptr.cast_mut().add(idx), amt);
+        }
+    }
+
+    /// Returns the length of this `String`.
+    ///
+    /// This length is in bytes, not [`char`]s or graphemes. In other words, it might not
+    /// be what a human considers the length of the string.
+    #[inline]
+    #[must_use]
+    pub fn len(&self) -> usize {
+        unsafe { sys::furi_string_size(self.0) }
+    }
+
+    /// Returns `true` if this `String` has a length of zero, and `false` otherwise.
+    #[inline]
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        unsafe { sys::furi_string_empty(self.0) }
+    }
+
+    /// Splits the string into two at the given byte index.
+    ///
+    /// Returns a newly allocated `String`. `self` contains bytes `[0, at)`, and the
+    /// returned `String` contains bytes `[at, len)`.
+    ///
+    /// Note that the capacity of `self` does not change.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `at` is beyond the last byte of the string.
+    #[inline]
+    #[must_use = "use `.truncate()` if you don't need the other half"]
+    pub fn split_off(&mut self, at: usize) -> String {
+        // SAFETY: Trimming the beginning of a C string results in a valid C string, as
+        // long as the nul byte is not trimmed.
+        assert!(at <= self.len());
+        let ret = String(unsafe { sys::furi_string_alloc_set_str(self.as_c_ptr().add(at)) });
+        self.truncate(at);
+        ret
+    }
+
+    /// Truncates this `String`, removing all contents.
+    ///
+    /// While this means the `String` will have a length of zero, it does not touch its
+    /// capacity.
+    #[inline]
+    pub fn clear(&mut self) {
+        unsafe { sys::furi_string_reset(self.0) };
+    }
+}
+
+impl Clone for String {
+    fn clone(&self) -> Self {
+        Self(unsafe { sys::furi_string_alloc_set(self.0) })
+    }
+}
+
+impl Default for String {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl AsRef<CStr> for String {
+    #[inline]
+    fn as_ref(&self) -> &CStr {
+        self.as_c_str()
+    }
+}
+
+impl From<char> for String {
+    fn from(value: char) -> Self {
+        let mut buf = String::with_capacity(value.len_utf8());
+        buf.push(value);
+        buf
+    }
+}
+
+impl From<&str> for String {
+    fn from(value: &str) -> Self {
+        let mut buf = String::with_capacity(value.len());
+        buf.push_str(value);
+        buf
+    }
+}
+
+impl From<&mut str> for String {
+    fn from(value: &mut str) -> Self {
+        From::<&str>::from(value)
+    }
+}
+
+impl From<&CStr> for String {
+    fn from(value: &CStr) -> Self {
+        Self(unsafe { sys::furi_string_alloc_set_str(value.as_ptr()) })
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl From<Box<str>> for String {
+    fn from(value: Box<str>) -> Self {
+        String::from(value.as_ref())
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> From<Cow<'a, str>> for String {
+    fn from(value: Cow<'a, str>) -> Self {
+        String::from(value.as_ref())
+    }
+}
+
+impl Extend<String> for String {
+    fn extend<T: IntoIterator<Item = String>>(&mut self, iter: T) {
+        iter.into_iter().for_each(move |s| self.push_string(&s));
+    }
+}
+
+impl Extend<char> for String {
+    fn extend<T: IntoIterator<Item = char>>(&mut self, iter: T) {
+        let iterator = iter.into_iter();
+        let (lower_bound, _) = iterator.size_hint();
+        self.reserve(lower_bound);
+        iterator.for_each(move |c| self.push(c));
+    }
+}
+
+impl<'a> Extend<&'a char> for String {
+    fn extend<T: IntoIterator<Item = &'a char>>(&mut self, iter: T) {
+        self.extend(iter.into_iter().cloned());
+    }
+}
+
+impl<'a> Extend<&'a str> for String {
+    fn extend<T: IntoIterator<Item = &'a str>>(&mut self, iter: T) {
+        iter.into_iter().for_each(move |s| self.push_str(s));
+    }
+}
+
+impl<'a> Extend<&'a CStr> for String {
+    fn extend<T: IntoIterator<Item = &'a CStr>>(&mut self, iter: T) {
+        iter.into_iter().for_each(move |s| self.push_c_str(s));
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl Extend<Box<str>> for String {
+    fn extend<T: IntoIterator<Item = Box<str>>>(&mut self, iter: T) {
+        iter.into_iter().for_each(move |s| self.push_str(&s));
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> Extend<Cow<'a, str>> for String {
+    fn extend<T: IntoIterator<Item = Cow<'a, str>>>(&mut self, iter: T) {
+        iter.into_iter().for_each(move |s| self.push_str(&s));
+    }
+}
+
+impl FromIterator<String> for String {
+    fn from_iter<T: IntoIterator<Item = String>>(iter: T) -> Self {
+        let mut iterator = iter.into_iter();
+
+        // Because we're iterating over `String`s, we can avoid at least one allocation by
+        // getting the first string from the iterator and appending to it all the
+        // subsequent strings.
+        match iterator.next() {
+            None => String::new(),
+            Some(mut buf) => {
+                buf.extend(iterator);
+                buf
+            }
+        }
+    }
+}
+
+impl FromIterator<char> for String {
+    fn from_iter<T: IntoIterator<Item = char>>(iter: T) -> Self {
+        let mut buf = String::new();
+        buf.extend(iter);
+        buf
+    }
+}
+
+impl<'a> FromIterator<&'a char> for String {
+    fn from_iter<T: IntoIterator<Item = &'a char>>(iter: T) -> Self {
+        let mut buf = String::new();
+        buf.extend(iter);
+        buf
+    }
+}
+
+impl<'a> FromIterator<&'a str> for String {
+    fn from_iter<T: IntoIterator<Item = &'a str>>(iter: T) -> Self {
+        let mut buf = String::new();
+        buf.extend(iter);
+        buf
+    }
+}
+
+impl<'a> FromIterator<&'a CStr> for String {
+    fn from_iter<T: IntoIterator<Item = &'a CStr>>(iter: T) -> Self {
+        let mut buf = String::new();
+        buf.extend(iter);
+        buf
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl FromIterator<Box<str>> for String {
+    fn from_iter<T: IntoIterator<Item = Box<str>>>(iter: T) -> Self {
+        let mut buf = String::new();
+        buf.extend(iter);
+        buf
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> FromIterator<Cow<'a, str>> for String {
+    fn from_iter<T: IntoIterator<Item = Cow<'a, str>>>(iter: T) -> Self {
+        let mut buf = String::new();
+        buf.extend(iter);
+        buf
+    }
+}
+
+impl Add<&str> for String {
+    type Output = String;
+
+    #[inline]
+    fn add(mut self, rhs: &str) -> Self::Output {
+        self.push_str(rhs);
+        self
+    }
+}
+
+impl AddAssign<&str> for String {
+    #[inline]
+    fn add_assign(&mut self, rhs: &str) {
+        self.push_str(rhs);
+    }
+}
+
+impl PartialEq for String {
+    fn eq(&self, other: &Self) -> bool {
+        unsafe { sys::furi_string_equal(self.0, other.0) }
+    }
+}
+
+impl PartialEq<CStr> for String {
+    fn eq(&self, other: &CStr) -> bool {
+        unsafe { sys::furi_string_equal_str(self.0, other.as_ptr()) }
+    }
+}
+
+impl PartialEq<String> for CStr {
+    fn eq(&self, other: &String) -> bool {
+        other.eq(self)
+    }
+}
+
+impl PartialEq<str> for String {
+    fn eq(&self, other: &str) -> bool {
+        self.to_bytes().eq(other.as_bytes())
+    }
+}
+
+impl PartialEq<String> for str {
+    fn eq(&self, other: &String) -> bool {
+        other.eq(self)
+    }
+}
+
+impl PartialEq<&str> for String {
+    fn eq(&self, other: &&str) -> bool {
+        self.eq(*other)
+    }
+}
+
+impl PartialEq<String> for &str {
+    fn eq(&self, other: &String) -> bool {
+        other.eq(*self)
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl PartialEq<CString> for String {
+    fn eq(&self, other: &CString) -> bool {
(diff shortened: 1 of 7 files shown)


crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// capacity.
+    pub fn clear(&mut self) {
+        unsafe { sys::furi_string_reset(self.0) };
+    }
+}
+
+/// A Furi string reference.
+///
+/// This is similar to Rust's [`str`], but backed by a `FuriString`.
+pub struct Str(sys::FuriString);
```
str4d, line 95: This type needs some work. Ideally it would be a dynamically-sized type (ensuring that the compiler enforces it is only ever used by reference), but the typical way I know of to do this is to embed a ZST slice `[()]`, but `&[()]` is a fat pointer and we want to cast to / from `*const sys::FuriString` which is a thin pointer.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,259 @@
...
+};
+
+#[cfg(feature = "alloc")]
+use alloc::ffi::CString;
+
+use flipperzero_sys as sys;
+
+/// A Furi string.
+///
+/// This is similar to Rust's [`String`], but used inside the Flipper Zero SDK.
```
dcoles, line 17: I think it's actually closer to `CString` as it isn't guaranteed to be UTF-8 encoded and is always `nul` terminated.

crates/flipperzero/src/furi/string.rs
dcoles: We could also use `core::CStr` rather than making our own reference type. The disadvantage of this is that we would then have to pull in the Rust implementations of `len`, `eq`, `cmp`, etc. rather than piggybacking off the C-implementation.

str4d: The problem with reusing `CStr` is that then we couldn't use `Str` with any Furi API that expects `*const sys::FuriString`, which is my goal here.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,259 @@
...
+            _ => unsafe { sys::furi_string_utf8_push(self.0, ch as u32) },
+        }
+    }
+
+    /// Shortens this `String` to the specified length.
+    ///
+    /// If `new_len` is greater than the string's current length, this has no effect.
+    pub fn truncate(&mut self, new_len: usize) {
+        if new_len < self.len() {
+            unsafe { sys::furi_string_right(self.0, new_len) };
```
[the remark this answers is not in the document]
str4d: Well that's useful to know, because the documentation for `furi_string_right` strongly suggests they are the other way round:
> Return the string right truncated from the 'index' position to the last position

crates/flipperzero/src/furi/string.rs
str4d: Yeah, I originally wrote that, but Rust's `CString` is a very minimal type, whereas here I'm trying to expose a similar interface to `String` (given that the Flipper Zero SDK exposes all that functionality). The `nul` termination in particular doesn't really matter in this context, because it is handled for you (vs when using a `CString` you explicitly need to care).

I guess it can be described as "This is similar to Rust's `CString`, but with the flexibility of Rust's `String`". I'll think of something to write here.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn push_str(&mut self, string: &str) {
+        self.reserve(string.len());
+        for ch in string.chars() {
+            self.push(ch);
+        }
+    }
+
+    /// Reserves capacity for at least `additional` bytes more than the current length.
+    pub fn reserve(&mut self, additional: usize) {
+        unsafe { sys::furi_string_reserve(self.0, self.len() + additional) };
```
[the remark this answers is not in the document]
str4d: Thanks. The documentation of it was unclear, and I read it as meaning that the `nul` character was handled for you (since every other method appears to handle the `nul`).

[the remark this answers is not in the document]
str4d: Wait, hang on. Doesn't that mean `sys::furi_string_size` returns the length including the `nul` character? In which case this method is fine, and the bug is that `String::len` requires an additional `- 1`?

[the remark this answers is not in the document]
str4d: Okay, [this part of the contract check](https://sourcegraph.com/github.com/P-p-H-d/mlib@62c8ac3e5d4a7a4f8757328e7a80286fde2686b6/-/blob/m-string.h?L40-41) appears to indicate that `string_size` returns the length of the string up to, but not including, the `nul` character.

And [this part of `string_reserve`](https://sourcegraph.com/github.com/P-p-H-d/mlib@62c8ac3e5d4a7a4f8757328e7a80286fde2686b6/-/blob/m-string.h?L318-322) ensures that the requested `alloc` is at least `size + 1`, which confirms that `alloc` needs to count the `nul` character.

crates/flipperzero/src/furi/string.rs
dcoles: In those cases, wouldn't you use `&furi::String` rather than a new `&furi::Str` type?

Both `&str` and `&CStr` are both just slices of `u8` and `c_char` (with the guarantee of being `nul`-terminated) respectively. Just holding a `&str` doesn't mean that it's backed by a `String` or that you can cast between them. There's not an equivalent `FuriStringView` as far as I can tell.

Though `FuriString` should implement `Borrow/AsRef<CStr>` as we can always get a `&CStr` view on the backing data.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,259 @@
...
+#[cfg(feature = "alloc")]
+use alloc::ffi::CString;
+
+use flipperzero_sys as sys;
+
+/// A Furi string.
+///
+/// This is similar to Rust's [`String`], but used inside the Flipper Zero SDK.
+#[derive(Eq)]
+pub struct String(*mut sys::FuriString);
```
dcoles, line 19: Looking at [`furi/core/string.h`](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware/-/blob/furi/core/string.h) and the underlying [m-string](https://github.com/P-p-H-d/mlib/tree/master#m-string), I think we can safely define it as `String(sys::FuriString)` so we can take advantage of m-string's ability to store small strings entirely on the stack.

crates/flipperzero/src/furi/string.rs
dcoles: Sort of related, I've been wondering where the best place to put a `FuriString` Rust type would be.

There's a few fundamental types, like this one, that would be good to have safe wrappers for low level use (e.g. in `flipperzero-rt`'s panic handler). So far I've been stuffing these into `flipperzero-sys`, but perhaps this warrants a `flipperzero-core` for the lowest level of safe wrappers. Then `flipperzero` could provide a wrapper type with stronger guarantees (e.g. is always valid UTF-8).

crates/flipperzero/src/furi/string.rs
str4d: I thought about the `Str` type more, and realised that if it is going to be an analogue of `str`, and in particular be a "string slice" that can be sub-sliced, then this means:

- It cannot be used with any Furi API that expects `*const sys::FuriString` (as sub-slices would be of the data within a `FuriString`, and passing the full string would be semantically invalid).
- It cannot include the nul terminator like `CStr` (because only sub-slices that preserve the end of the slice could do this).

So the type that can be passed into Flipper Zero APIs requiring  `*const sys::FuriString` will just be `&String`, and `Str` would effectively just be a wrapper around `&[u8]`. There might still be some use for `&Str` (as a borrowed `FuriString`, just as `&str` is a borrowed Rust string and `&CStr` is a borrowed C string), but only on the Rust side.

crates/flipperzero/src/furi/string.rs
str4d: I don't think that helps? [`sys::furi_string_alloc()` always allocates a `FuriString` on the heap](https://github.com/flipperdevices/flipperzero-firmware/blob/c1bb10a6947f56e7cb71912b864b6d2b969f6155/furi/core/string.c#L25-L29), regardless of how the `string_t` underneath is allocated.

The Flipper Zero SDK does not expose the `FuriString` C struct directly; the `sys::FuriString` type is being generated by `bindgen` as a target for the `*mut FuriString` pointer. If we define this as `String(sys::FuriString)` then we'd effectively only be able to represent `&mut String` and `&String`, not an owned type.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,261 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if `at` is beyond the last byte of the string.
+    #[inline]
+    #[must_use = "use `.truncate()` if you don't need the other half"]
+    pub fn split_off(&mut self, at: usize) -> String {
+        // SAFETY: Trimming the beginning of a C string results in a valid C string, as
+        // long as the nul byte is not trimmed.
+        assert!(at <= self.len());
```
str4d, line 160: [`String::split_off`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.split_off) similarly panics if this condition is not met, as well as if `at` is not at a `char` boundary in the UTF-8 encoded string. For now I've dropped the UTF-8 panic (given that `FuriString` doesn't try to preserve UTF-8 encoding) but kept the OOB panic as-is (there are plenty of similar panics we can't avoid due to the Flipper Zero SDK not exposing error handling for these methods).

If we want to avoid the panic then we could instead make this `try_split_off(&mut self, at: usize) -> Result<String, SplitOffError>`.

crates/flipperzero/src/furi/string.rs
dcoles: That's correct. `sys::furi_string_alloc` always allocates a `FuriString` on the heap, but there's nothing stopping you from putting an uninitialized `FuriString` on the stack and then calling `string_init(&furi_string.string)` on that structure (other than there being no `sys::furi_string_init`—I'll see if they'll accept a PR).

dcoles: Oh, I see what you mean. It's an opaque type, so it's not possible to even know `sizeof(FuriString)`.

Still, I'll see if it might be possible to expose that detail.

crates/flipperzero/src/furi/string.rs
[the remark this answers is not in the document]
dcoles: I hate looking at off-by-one issues like this. I got such a headache last night. 😄

crates/flipperzero/src/furi/string.rs
dcoles: That seems fine to me. 👍

crates/flipperzero/src/furi/string.rs
dcoles: Filed a feature request on the SDK: https://github.com/flipperdevices/flipperzero-firmware/issues/2461

str4d: In any case, we don't need to block on this, as `furi::String` is an opaque type, so we can change its internals whenever we want.

crates/flipperzero/src/furi/string.rs
str4d: I put this under `flipperzero::furi` because that's where the other Furi wrappers were. Happy to move it elsewhere. Another idea: a `flipperzero-furi` crate, that can then be used directly by e.g. `flipperzero-rt` and be re-exported as `flipperzero::furi`?

> Then `flipperzero` could provide a wrapper type with stronger guarantees (e.g. is always valid UTF-8).

Something I noticed while working on this PR is that `String` and `str` have diverse APIs, while `CString` and `CStr` are quite minimal. For `CStr` it makes sense: you can't treat it like a string slice because C strings must always be nul-terminated. But `CString` has confused me; I guess there hasn't been much motivation to extend its API given that most Rust programs will just use `String`.

I think the direct `FuriString` wrapper should be expressive (as I've been implementing), but we'd also want the UTF-8 wrapper to be similarly usable (perhaps even more closely aligning with Rust `String`). If we did introduce this kind of wrapper, then we'd need to make it opaque without a `Deref` impl, to ensure that the "always valid UTF-8" invariant is maintained.

str4d:
The PR currently focuses on providing newly-constructed, Rust-owned `FuriString`s. But we will also want the ability to represent `FuriString`s inside types owned by the Flipper Zero SDK (which we cannot free in a Rust `Drop` impl). I think the way to handle that is to add a constructor like:
```rust
impl String {
    pub(crate) unsafe fn from_borrowed_mut<'s>(raw: *mut FuriString) -> &'s mut String {
        todo!("pointer magic")
    }
}
```
which is only used from inside other wrapper types, enabling them to implement APIs like:
```rust
pub struct OtherWrapper(*mut sys::SomeType);

impl OtherWrapper {
    pub fn name_mut(&mut self) -> &mut furi::String {
        unsafe { furi::String::from_borrowed_mut((*self.0).name) }
    }
}
```

crates/flipperzero/src/furi/string.rs
dcoles: > Another idea: a `flipperzero-furi` crate, that can then be used directly by e.g. `flipperzero-rt` and be re-exported as `flipperzero::furi`?

That's not a bad idea either. I think for the time being I'll just keep using `flipperzero-sys` for these low-level wrappers (the [`windows`](https://crates.io/crates/windows) crate [does something similar](https://github.com/microsoft/windows-rs/blob/master/crates/libs/sys/src/core/mod.rs)).

> But `CString` has confused me; I guess there hasn't been much motivation to extend its API given that most Rust programs will just use `String`.

Right. I really see it mostly as an intermediary for working with C APIs. A bit of a necessary evil.

The reason I do think having a `sys::FuriString` wrapper (with/without UTF-8 guarantee) is useful here is because the device is fairly memory constrained, so people might actually care about unnecessary allocations. I like your approach of an expressive `furi::String` type with a UTF-8 guarantee since it provides a far more friendly API than `CString`, while still being quite careful about resources.

I imagine a `sys::furi::String` low-level type would feel a bit more like `Vec<c_char>` that exposed the `FuriString` API, but not much else.

crates/flipperzero/src/furi/string.rs
dcoles: I agree. Let's move forward with the current implementation.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/8eec33a50d11e5b994f354cb756f6b7e56faf45c..d16eb98cb920c7c199c44e07c055a163e45a56bd) to bring in the test harness.

str4d:
Pushed a test suite (adapted from https://github.com/rust-lang/rust/blob/master/library/alloc/tests/string.rs). The uncommented tests are the ones for APIs I have currently implemented. The tests do all pass, but I've seen a few occasional panics that I haven't been able to pin down yet.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/97814b77036fab45c3478ee0837f24a002a1dd06..25a1dd76c36f21194477ede5f5f7d9c4aef40872) to add some more trait impls.

str4d:
Almost all the remaining APIs of `std::string::String` that I haven't implemented here yet depend in some way on whether we want `flipperzero::furi::string::String` to be:
- a `CString` with better APIs?
- a `String` that allows non-UTF-8?
- a `String` that should enforce UTF-8 (and thus enable `&str` to be used for slicing) but happens to have a trailing `nul` in its internal representation?

In particular, I'm trying to decide whether to provide a `String::chars` method. Rust's `char` type is defined as a Unicode scalar value, so while we can easily have `String::push(char)`, we cannot iterate over `chars` unless `String` is guaranteed to contain valid UTF-8. But most of the remaining APIs (`pop`, `remove`, `retain`, etc.) are implemented in terms of `String::chars`. So we either don't provide them, we restrict `String` to be UTF-8, or we provide them but in terms of bytes.

dcoles:
It's been a while since I've thought about this, but I think my direction of thinking was:
- Any `flipperzero::furi::string::String` must be trivially be converted into a `sys::FuriString` (and vice-versa)
  - This is to allow interacting with the FlipperZero API without needing additional allocations in either direction
- As such, `flipperzero::furi::string::String` is guaranteed to be `\0` terminated
- As such, it can implement `Borrow/AsRef<CStr>` but can't be trivially sliced 
- We can support things like `trim` or `mid` but they act in-place using the `furi_string_` methods
- We can support `to_bytes`/`to_bytes_with_null` to get `&[u8]` slices
- We can support `to_str` to get a `Result<&str, Utf8Error>`
- Where possible, you should still be able to use string literals (treated as utf-8 bytes)
  - e.g. `push(char)` or `push_str(&str)` are easy to implement
- It should be fine to allow accessing as a `&mut [u8]` as long as the slice excludes the final `\0`

I suspect we could even support an iterator over unicode scalars built on `furi_string_utf8_decode` that returned [`REPLACEMENT_CHARACTER`](https://doc.rust-lang.org/core/primitive.char.html#associatedconstant.REPLACEMENT_CHARACTER) or `None` whenever it encountered an invalid UTF-8 sequence, but I think we'd quickly see diminishing returns in respect to implementation effort to support the remaining methods that operate on `char`. At that point, it's probably best to just use a Rust `String` and convert it to `flipperzero::String` when you're done.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/25a1dd76c36f21194477ede5f5f7d9c4aef40872..b7c31b9a8231e66aaa20c3d6b95b113c7bc8719e) to fix merge conflict in `flipperzero::furi`.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/b7c31b9a8231e66aaa20c3d6b95b113c7bc8719e..6f884e1ef786e37d32bc345fcdc197dc632d035d) to rename `String::push_rust_str` back to `String::push_str` as the PR no longer tries to provide its own `str` type.

str4d:
I've [implemented](https://github.com/flipperzero-rs/flipperzero/compare/6f884e1ef786e37d32bc345fcdc197dc632d035d..0a6135588c089ce8fe656dc961273a4870027a85) a bunch of the pattern-based `str` methods now, as well as importing the corresponding test suite from `rust-lang/rust`.

At this point I suspect we are approaching diminishing returns. I can continue to slowly implement additional methods and get more of the tests uncommented and passing, but that doesn't necessarily have to block this PR. I'm going to pause that work and ensure that I can use `furi::String` in #47.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,784 @@
...
+/// nul-terminated string with no nul bytes in the middle. It also has additional methods
+/// to provide the flexibility of Rust's [`String`]. It is used in various APIs of the
+/// Flipper Zero SDK.
+///
+/// This type does not requre the `alloc` feature flag, because it does not use the Rust
+/// allocator. Very short strings (7 bytes or fewer) are stored directly inside the
+/// `FuriString` struct (which is stored on the heap), while longer strings are allocated
+/// on the heap by the Flipper Zero firmware.
+#[derive(Eq)]
+pub struct String(*mut sys::FuriString);
```
dcoles, line 42: I'm still wondering if we should explicitly call this `FuriString` or just `String`.

`String` is nice since, if we included it in the prelude, things almost look like regular Rust (`String::from("word")` does what you expect, common functions use `String`, etc.). However, I think this leads to more confusion in the long term (accidentally looking at the `std::string::String`  docs, wondering why a library API that takes a `String` doesn't work, not being clear whether an API takes a standard Rust `String` or a Furi `String`, etc.).

I'm leaning towards `FuriString` at the moment.

dcoles (approved):

dcoles:
Looking great! My vote is to get the current work merged in (it's already far superior to `CString`) and add any additional functionality on a case-by-case basis.

crates/flipperzero/src/furi/string.rs
str4d: Originally I was hoping for this to actually be a drop-in replacement for `String`. Now that it isn't the case, I agree with renaming it for clarity.

crates/flipperzero/src/furi/string.rs
```diff
@@ -0,0 +1,784 @@
...
+            f.write_char(byte as char)?;
+        }
+        f.write_char('"')
+    }
+}
+
+impl fmt::Display for String {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        for byte in self.to_bytes().escape_ascii() {
```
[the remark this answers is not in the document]
str4d: Turns out this was easier to implement than I thought it would be. Done!

str4d:
Okay, I think that's everything 🙂

str4d:
Okay, having committed the new file, _now_ that's everything 😂

crates/flipperzero/src/furi/string.rs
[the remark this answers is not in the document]
dcoles: Nice! That's awesome!

dcoles:
Thanks for the PR! This one was big one and will be extremely useful!

dcoles:
@str4d Since you've by far the most active contributor to this repository, I've invited you to @flipperzero-rs/maintainers. This should grant you fairly broad permission to push, merge, etc. No obligation to accept, but it would be good to have another pair of eyes watching things since I'm not always as active as I would like.

PR #55, opened 2023-03-13
shell.nix: add shell.nix

twitchyliquid64:
This gives a ready-to-go Nix environment for development. Nix users can just run `nix-shell` and be ready to go.

 * A nightly Rust compiler, configured to build `thumbv7em-none-eabihf`
 * "The good kush" in terms of rust extensions
 * Python deps you need to run any official firmware tooling
 * System dependencies to run any of the flipperzero-rs tooling

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,46 @@
+let
+  here = toString ./.;
+  moz_overlay = import (builtins.fetchTarball
+    "https://github.com/mozilla/nixpkgs-mozilla/archive/master.tar.gz");
+  pkgs = import <nixpkgs> { overlays = [ moz_overlay ]; };
+  rust = (pkgs.rustChannelOf {
+    channel = "nightly";
+  }).rust.override {
+    extensions = [ "rust-src" "rust-analysis" ];
+    targets = [ "thumbv7em-none-eabihf" ];
+  };
+  rustPlatform = pkgs.makeRustPlatform {
+    rustc = rust;
+    cargo = rust;
+  };
+
+
+
+  my-python-packages = p: with p; [
+    pyserial
+  ];
+
+  systemDeps = with pkgs; [
+    udev
+  ];
+  systemLibStr = pkgs.lib.makeLibraryPath systemDeps;
+
+in pkgs.mkShell {
+  packages = [
+    (pkgs.python3.withPackages my-python-packages)
+  ];
+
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -0,0 +1,46 @@
+let
+  here = toString ./.;
+  moz_overlay = import (builtins.fetchTarball
+    "https://github.com/mozilla/nixpkgs-mozilla/archive/master.tar.gz");
+  pkgs = import <nixpkgs> { overlays = [ moz_overlay ]; };
+  rust = (pkgs.rustChannelOf {
+    channel = "nightly";
```
dcoles, line 7: Can you use `rustToolchain = "crates/rust-toolchain.toml"` so this follows the supported nightly toolchain?

shell.nix
```diff
@@ -0,0 +1,46 @@
...
+  };
+  rustPlatform = pkgs.makeRustPlatform {
+    rustc = rust;
+    cargo = rust;
+  };
+
+
+
+  my-python-packages = p: with p; [
+    pyserial
```
dcoles, line 20: As of #54 there should be much less of a need for Python, but there is still [`crates/build-examples.py`](https://github.com/flipperzero-rs/flipperzero/blob/48392243271f7b9151050ed5f79feca621a5f26b/crates/build-examples.py) helper.

I believe most Python installations are about 500 MB in size. Do we have to be wary about size at all?

shell.nix
```diff
@@ -0,0 +1,46 @@
...
+  systemLibStr = pkgs.lib.makeLibraryPath systemDeps;
+
+in pkgs.mkShell {
+  packages = [
+    (pkgs.python3.withPackages my-python-packages)
+  ];
+
+  buildInputs = [
+    rust
+    pkgs.gcc-arm-embedded
```
dcoles, line 35: These crates don't have any direct dependency on `gcc` (though we do need to use the public Flipper Zero toolchain for the headers provided in the toolchain's sysroot). As such, do we need to pull ARM GCC into the environment?

dcoles:
Hi @twitchyliquid64,

This looks like a small enough addition, so I don't mind merging it. However, I don't have much experience with Nyx, so please bear with my ignorant questions.

dcoles:
Closing pull request, since it hasn't had any activity for ~5 months.

PR #57, opened 2023-03-17
Test harness

### LINKED ISSUE #53: Test harness
Currently all testing facilities are disabled for the crates in this repo. It makes sense because everything will only run on an actual Flipper Zero, but it does make verification of things tricky.

It would be useful to have a test harness, that runs the test binaries on a connected Flipper Zero. I'm imagining something like the [integration tests for the `yubikey` crate](https://github.com/iqlusioninc/yubikey.rs/blob/main/tests/integration.rs), which require setting a `YUBIKEY_SERIAL` environment variable in order to run. The [`test-with`](https://crates.io/crates/test-with) crate might also be helpful here, specifically `#[test_with::env(FLIPPER_ZERO_SERIAL_PORT)]` or something, that can be used to enable at compilation time tests that require a connected Flipper Zero.

str4d:
Closes #53.

### DIFF
--- a/crates/.cargo/config.toml
+++ b/crates/.cargo/config.toml
@@ -1,4 +1,5 @@
 [target.thumbv7em-none-eabihf]
+runner = "python3 ../cargo-runner.py"
 rustflags = [
     # CPU is Cortex-M4 (STM32WB55)
     "-C", "target-cpu=cortex-m4",
--- a/crates/cargo-runner.py
+++ b/crates/cargo-runner.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+# Helper script for running binaries on a connected Flipper Zero.
+
+import argparse
+import os
+import sys
+from pathlib import Path
+from subprocess import run
+
+TOOLS_PATH = '../tools'
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('binary', type=Path)
+    parser.add_argument('arguments', nargs=argparse.REMAINDER)
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    # Run the given FAP binary on a connected Flipper Zero.
+    result = run(
+        [
+            'cargo',
+            'run',
+            '--quiet',
+            '--release',
+            '--bin',
+            'run-fap',
+            '--',
+            os.fspath(args.binary),
+        ] + args.arguments,
+        cwd=os.path.join(os.path.dirname(__file__), TOOLS_PATH),
+    )
+    if result.returncode:
+        sys.exit(result.returncode)
+
+
+if __name__ == '__main__':
+    main()
--- a/crates/flipperzero/src/lib.rs
+++ b/crates/flipperzero/src/lib.rs
@@ -20,4 +20,7 @@ pub mod __internal {
     pub use ufmt;
 }
 
-flipperzero_test::tests_runner!([]);
+flipperzero_test::tests_runner!(
+    name = "flipperzero-rs Unit Tests",
+    [crate::furi::message_queue::tests, crate::furi::sync::tests]
+);
--- a/crates/flipperzero/src/lib.rs
+++ b/crates/flipperzero/src/lib.rs
@@ -1,6 +1,7 @@
 //! High-level bindings for the Flipper Zero.
 
 #![no_std]
+#![cfg_attr(test, no_main)]
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
@@ -18,3 +19,5 @@ pub mod __internal {
     // Re-export for use in macros
     pub use ufmt;
 }
+
+flipperzero_test::tests_runner!([]);
--- a/crates/test/macros/src/deassert.rs
+++ b/crates/test/macros/src/deassert.rs
@@ -1,0 +1,101 @@
+use quote::quote;
+use syn::{parse, Block, Expr, ExprMacro, ExprTuple, Stmt};
+
+/// Find and replace macro assertions inside the given block with `return Err(..)`.
+///
+/// The following assertion macros are replaced:
+/// - [`assert`]
+/// - [`assert_eq`]
+/// - [`assert_ne`]
+pub(crate) fn box_block(mut block: Box<Block>) -> parse::Result<Box<Block>> {
+    block.stmts = block_stmts(block.stmts)?;
+    Ok(block)
+}
+
+/// Searches recursively through block statements to find and replace macro assertions
+/// with `return Err(..)`.
+///
+/// The following assertion macros are replaced:
+/// - [`assert`]
+/// - [`assert_eq`]
+/// - [`assert_ne`]
+fn block_stmts(stmts: Vec<Stmt>) -> parse::Result<Vec<Stmt>> {
+    stmts
+        .into_iter()
+        .map(|stmt| match stmt {
+            Stmt::Expr(Expr::Block(mut e)) => {
+                e.block.stmts = block_stmts(e.block.stmts)?;
+                Ok(Stmt::Expr(Expr::Block(e)))
+            }
+            Stmt::Expr(Expr::Macro(m)) => expr_macro(m).map(Stmt::Expr),
+            Stmt::Semi(Expr::Macro(m), trailing) => expr_macro(m).map(|m| Stmt::Semi(m, trailing)),
+            _ => Ok(stmt),
+        })
+        .collect::<Result<_, _>>()
+}
+
+/// Replaces macro assertions with `return Err(..)`.
+///
+/// The following assertion macros are replaced:
+/// - [`assert`]
+/// - [`assert_eq`]
+/// - [`assert_ne`]
+fn expr_macro(m: ExprMacro) -> parse::Result<Expr> {
+    if m.mac.path.is_ident("assert") {
+        let tokens = m.mac.tokens;
+        let tokens_str = tokens.to_string();
+        syn::parse(
+            quote!(
+                if !(#tokens) {
+                    return ::core::result::Result::Err(
+                        ::core::concat!("assertion failed: ", #tokens_str).into(),
+                    );
+                }
+            )
+            .into(),
+        )
+    } else if m.mac.path.is_ident("assert_eq") {
+        let (left, right) = binary_macro(m.mac.tokens)?;
+        let left_str = quote!(#left).to_string();
+        let right_str = quote!(#right).to_string();
+        syn::parse(
+            quote!(
+                if #left != #right {
+                    return ::core::result::Result::Err(
+                        ::flipperzero_test::TestFailure::AssertEq {
+                            left: #left_str,
+                            right: #right_str,
+                        }
+                    );
+                }
+            )
+            .into(),
+        )
+    } else if m.mac.path.is_ident("assert_ne") {
+        let (left, right) = binary_macro(m.mac.tokens)?;
+        let left_str = quote!(#left).to_string();
+        let right_str = quote!(#right).to_string();
+        syn::parse(
+            quote!(
+                if #left == #right {
+                    return ::core::result::Result::Err(
+                        ::flipperzero_test::TestFailure::AssertNe {
+                            left: #left_str,
+                            right: #right_str,
+                        }
+                    );
+                }
+            )
+            .into(),
+        )
+    } else {
+        Ok(Expr::Macro(m))
+    }
+}
+
+fn binary_macro(tokens: proc_macro2::TokenStream) -> parse::Result<(Expr, Expr)> {
+    let parts: ExprTuple = syn::parse(quote!((#tokens)).into())?;
+    assert_eq!(parts.elems.len(), 2);
+    let mut elems = parts.elems.into_iter();
+    Ok((elems.next().unwrap(), elems.next().unwrap()))
+}
--- a/crates/test/macros/src/lib.rs
+++ b/crates/test/macros/src/lib.rs
@@ -1,0 +1,303 @@
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::{quote, ToTokens};
+use syn::{
+    parse::{self, Parse},
+    punctuated::Punctuated,
+    spanned::Spanned,
+    token, Expr, ExprArray, Ident, Item, ItemMod, ReturnType, Stmt, Token,
+};
+
+mod deassert;
+
+struct TestRunner {
+    manifest_args: Punctuated<TestRunnerArg, Token![,]>,
+    test_suites: ExprArray,
+}
+
+impl Parse for TestRunner {
+    fn parse(input: parse::ParseStream) -> syn::Result<Self> {
+        let mut manifest_args = Punctuated::new();
+        if !input.peek(token::Bracket) {
+            loop {
+                if input.is_empty() || input.peek(token::Bracket) {
+                    break;
+                }
+                let value = input.parse()?;
+                manifest_args.push_value(value);
+                if input.is_empty() || input.peek(token::Bracket) {
+                    break;
+                }
+                let punct = input.parse()?;
+                manifest_args.push_punct(punct);
+            }
+        };
+
+        let test_suites = input.parse()?;
+
+        Ok(TestRunner {
+            manifest_args,
+            test_suites,
+        })
+    }
+}
+
+struct TestRunnerArg {
+    ident: Ident,
+    eq_token: Token![=],
+    value: Box<Expr>,
+}
+
+impl Parse for TestRunnerArg {
+    fn parse(input: parse::ParseStream) -> syn::Result<Self> {
+        let ident = input.parse()?;
+        let eq_token = input.parse()?;
+        let value = input.parse()?;
+        Ok(TestRunnerArg {
+            ident,
+            eq_token,
+            value,
+        })
+    }
+}
+
+#[proc_macro]
+pub fn tests_runner(args: TokenStream) -> TokenStream {
+    match tests_runner_impl(args) {
+        Ok(ts) => ts,
+        Err(e) => e.to_compile_error().into(),
+    }
+}
+
+fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
+    let TestRunner {
+        manifest_args,
+        test_suites,
+    } = syn::parse(args)?;
+
+    let test_suites = test_suites
+        .elems
+        .into_iter()
+        .map(|attr| {
+            let mut module = String::new();
+            for token in attr.to_token_stream() {
+                module.push_str(&token.to_string());
+            }
+            let module = module.trim_start_matches("crate::");
+
+            (
+                quote!(#attr::__test_list().len()),
+                quote!(#attr::__test_list().map(|(name, test_fn)| (#module, name, test_fn))),
+            )
+        })
+        .collect::<Vec<_>>();
+
+    let test_counts = test_suites.iter().map(|(count, _)| count);
+    let test_lists = test_suites.iter().map(|(_, list)| list);
+
+    let manifest_args = manifest_args.into_iter().map(
+        |TestRunnerArg {
+             ident,
+             eq_token,
+             value,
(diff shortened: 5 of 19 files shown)


crates/.cargo/config.toml
```diff
@@ -1,4 +1,5 @@
 [target.thumbv7em-none-eabihf]
+runner = "../tools/cargo-runner.sh"
```
str4d, line 2: This works reliably from inside the `crates/` directory, but I have not tested from elsewhere yet.

crates/flipperzero/src/lib.rs
```diff
@@ -16,3 +17,56 @@ pub mod __internal {
     // Re-export for use in macros
     pub use ufmt;
 }
+
+// Infrastructure for running unit tests.
```
str4d, line 21: This infrastructure will likely end up becoming a proc macro (so it can be used for both unit tests and integration tests), but for now I'm implementing it explicitly to make its actions clear.

tools/src/bin/run-fap.rs
```diff
@@ -0,0 +1,109 @@
...
+        .send_file(&cli.fap, &dest_file)
+        .map_err(Error::FailedToUploadFap)?;
+
+    let cli = store.cli_mut();
+
+    // Wait for no application to be running.
+    wait_for_idle(cli)?;
+
+    // Run the FAP.
+    cli.send_and_wait_eol(&format!("loader open Applications {}\r", dest_file))?;
```
str4d, line 95: Currently it is not possible to provide arguments to external FAPs. I've opened https://github.com/flipperdevices/flipperzero-firmware/issues/2505 upstream requesting this. In the meantime, this just means that test filtering won't work.

crates/flipperzero/src/lib.rs
str4d: I've now made it a proc macro, because it ended up being easier (as then I could use a second proc macro to process the tests themselves into a form that the test runner can use).

crates/test/macros/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+                println!("Leaked: {}\r", heap_before - heap_after);
+
+                // Final Report
+                if failed == 0 {
+                    println!("Status: PASSED\r");
+                    0
+                } else {
+                    println!("Status: FAILED\r");
+                    1
+                }
```
str4d, line 165: Currently the test runner prints the same information as the [`flipperzero-firmware` unit tests](https://github.com/flipperdevices/flipperzero-firmware/blob/0.78.1/documentation/UnitTests.md) in the device logs. However, for the purpose of `cargo test` we need to extract this information back to the host. I currently believe this means we'll need to save the results to a file and then download it before cleaning up the device.

crates/test/macros/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+            // Test runner entry point
+            fn main(_args: *mut u8) -> i32 {
+                let heap_before = unsafe { sys::memmgr_get_free_heap() };
+                let cycle_counter = unsafe { sys::furi_get_tick() };
+
+                let failed = 0;
+                for test in test_list() {
+                    // When the test assertions fail, we can't catch the panics.
+                    // TODO: Replace panics with Results.
+                    test();
```
str4d, line 141: The current state of this PR does successfully run tests on the Flipper Zero. However, if any test assertion fails, the unit test FAP crashes so we won't get any results. Given how `flipperzero_rt` intercepts the panic handler, and `catch_unwind` isn't available anyway without `std`, I think the best solution is to require tests to return `Result<(), &'static str>`, and then maybe use the proc macro to rewrite the assertion-based tests if possible.

str4d: Intercepting the `assert` macros works very nicely!

crates/test/macros/src/deassert.rs
```diff
@@ -0,0 +1,99 @@
+use quote::quote;
+use syn::{parse, Block, Expr, ExprMacro, ExprTuple, Stmt};
+
+/// Find and replace macro assertions inside the given block with `return Err(..)`.
+///
+/// The following assertion macros are replaced:
+/// - [`assert`]
+/// - [`assert_eq`]
+/// - [`assert_ne`]
```
str4d, line 9: In addition to these macros, we should also try to hunt for and replace `.unwrap()` and `.expect(..)` with something that similarly causes `Err(..)` to be returned. That being said, I'm fine with this being best-effort; the goal is to make writing unit tests easy, not to completely erase knowledge of the fact that panics will break `cargo test`.

tools/src/bin/run-fap.rs
```diff
@@ -0,0 +1,111 @@
...
+    let port = serialport::new(&port_info.port_name, serial::BAUD_115200)
+        .timeout(Duration::from_secs(30))
+        .open()
+        .map_err(Error::FailedToOpenSerialPort)?;
+    let mut store = storage::FlipperStorage::new(port);
+    store.start().map_err(Error::FailedToStartSerialInterface)?;
+
+    // Upload the FAP to a temporary directory.
+    let dest_dir =
+        storage::FlipperPath::from(format!("/ext/tmp-{:08x}", thread_rng().gen::<u32>()));
```
str4d, line 82: I haven't yet figured out if any of the Flipper Zero SDKs let a running external FAP learn where it was loaded from. Without that knowledge, it's gonna be tricky to reliably redirect stdout to a file that both won't collide with an existing file on the Flipper Zero, and that this tool can find.

str4d:
Okay, I have this working on Linux and macOS! Current output:
```
$ cd crates/
$ cargo test
   Compiling proc-macro2 v1.0.52
   Compiling unicode-ident v1.0.8
   Compiling quote v1.0.26
   Compiling syn v1.0.109
   Compiling ufmt-write v0.1.0
   Compiling flipperzero-rt v0.7.2 (/path/to/flipperzero/crates/rt)
   Compiling ufmt-macros v0.3.0
   Compiling flipperzero-test-macros v0.1.0 (/path/to/flipperzero/crates/test/macros)
   Compiling ufmt v0.2.0
   Compiling flipperzero-sys v0.7.2 (/path/to/flipperzero/crates/sys)
   Compiling flipperzero-test v0.1.0 (/path/to/flipperzero/crates/test)
   Compiling flipperzero-alloc v0.7.2 (/path/to/flipperzero/crates/alloc)
   Compiling flipperzero v0.7.2 (/path/to/flipperzero/crates/flipperzero)
    Finished test [optimized + debuginfo] target(s) in 4.73s
     Running unittests src/lib.rs (target/thumbv7em-none-eabihf/debug/deps/flipperzero-ec498257ef00abdb)

running 3 tests
test dolphin::tests::stats ... ok
test furi::message_queue::tests::capacity ... ok
test furi::sync::tests::unshared_mutex_does_not_block ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2ms
leaked: 88 bytes

$ # Unplug Flipper Zero
$ cargo test
    Finished test [optimized + debuginfo] target(s) in 0.01s
     Running unittests src/lib.rs (target/thumbv7em-none-eabihf/debug/deps/flipperzero-ec498257ef00abdb)
Error: unable to find Flipper Zero
error: test failed, to rerun pass `-p flipperzero --lib`

Caused by:
  process didn't exit successfully: `/path/to/flipperzero/crates/../tools/cargo-runner.sh /path/to/flipperzero/crates/target/thumbv7em-none-eabihf/debug/deps/flipperzero-ec498257ef00abdb` (exit status: 1)
```

tools/src/bin/run-fap.rs
str4d: I ended up just using a common fixed path of `/ext/flipperzero-rs-stdout`, on the assumption that external FAPs cannot run concurrently, so only one app at a time will ever be running, and as long as we immediately read the file and then remove it we're probably fine. This will do until the Flipper Zero SDK supports RPC usage.

crates/test/macros/src/lib.rs
str4d: Saving the results to a file is what I am doing now.

str4d:
And if I manually cause some tests to fail:
```
$ cargo test
   Compiling flipperzero v0.7.2 (/path/to/flipperzero/crates/flipperzero)
    Finished test [optimized + debuginfo] target(s) in 0.28s
     Running unittests src/lib.rs (target/thumbv7em-none-eabihf/debug/deps/flipperzero-ad4e99e6ec932a9a)

running 3 tests
test dolphin::tests::stats ... ok
test furi::message_queue::tests::capacity ... FAILED

---- furi::message_queue::tests::capacity stdout ----
assertion failed: queue.len() == 1

test furi::sync::tests::unshared_mutex_does_not_block ... FAILED

---- furi::sync::tests::unshared_mutex_does_not_block stdout ----
assertion failed: * value == 6


test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5ms
leaked: 88 bytes
```

Currently the test failures are printed in-line, to avoid requiring `alloc`.

tools/src/bin/run-fap.rs
JarvisCraft: You may try addressing the issues in Flipperzero [Telegram group](https://t.me/flipperzero), i.e. `Development (SW&HW)` channel.

str4d: https://github.com/flipperdevices/flipperzero-firmware/issues/2505#issuecomment-1474184238 indicated that an improved host-device communication protocol is planned for external FAPs, so that is likely what will be used instead when available.

str4d:
I've adjusted the test harness so it now also works for integration tests, and to allow the manifest arguments to be set (in particular, while adding more tests in subsequent local changes I need to bump the stack size).

There are probably more usability tweaks that could be made, but I think we'll discover those as the test harness is used, so I'm not planning to make any more changes to this PR other than in response to review.

crates/test/macros/src/deassert.rs
dcoles: That seems like a reasonable approach. Since we don't support `panic = "unwind"` (and [likely never will](https://github.com/flipperdevices/flipperzero-firmware/issues/1851)), I don't really see any other way of running tests than a) 1 binary per test, b) co-operative tests like you're doing here.

crates/test/macros/src/deassert.rs
```diff
@@ -0,0 +1,101 @@
...
+                }
+            )
+            .into(),
+        )
+    } else {
+        Ok(Expr::Macro(m))
+    }
+}
+
+fn binary_macro(tokens: proc_macro2::TokenStream) -> parse::Result<(Expr, Expr)> {
```
dcoles, line 96: What does this function do? It looks like it might be for unpacking a 2-argument macro into individual expressions.

dcoles (approved):

crates/.cargo/config.toml
str4d: It turns out that `target.<cfg>.runner` also works, so I might be able to specify a Windows script here as well with some effort.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/754d827f14ad03401bf0a8bdeaa0e9455439aab7..91452b59aca13dcc2f2b5802cd410413cfab23a4) to fix merge conflict with #59.

crates/.cargo/config.toml
str4d: I can't figure out an easy way to get this to work, so given that Unix platforms almost always have Python, and you're clearly already requiring it on Windows, let's just use that for now.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/91452b59aca13dcc2f2b5802cd410413cfab23a4..bab2e58543afd8d32c10081eaf576c0e803a9731) to replace `cargo-runner.sh` with `cargo-runner.py`. I've tested that this PR still works on Ubuntu and macOS; it should now also work on Windows (untested, but [Windows can execute `*.py` files directly if their extension is associated correctly](https://docs.python.org/3/faq/windows.html#how-do-i-make-python-scripts-executable)

crates/cargo-runner.py
```diff
@@ -0,0 +1,43 @@
...
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('binary', type=Path)
+    parser.add_argument('arguments', nargs=argparse.REMAINDER)
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    print()
```
dcoles, line 22: Why are you printing a newline here?

crates/cargo-runner.py
```diff
@@ -0,0 +1,43 @@
...
+        [
+            'cargo',
+            'run',
+            '--quiet',
+            '--release',
+            '--bin',
+            'run-fap',
+            '--',
+            os.fspath(args.binary),
+        ] + args.arguments,
```
dcoles, line 35: For something as simple as this script, we could probably even get away with just `+ sys.argv`.

dcoles:
@str4d It looks like the runner [might be causing problems for GitHub Actions](https://github.com/flipperzero-rs/flipperzero/actions/runs/4581112299/jobs/8090398886?pr=57). Can you take a look?

dcoles:
> it should now also work on Windows (untested, but [Windows can execute *.py files directly if their extension is associated correctly](https://docs.python.org/3/faq/windows.html#how-do-i-make-python-scripts-executable)).

Looks like `runner` must be an application, not just an associated file type:

```
PS> ./cargo-runner.py
usage: cargo-runner.py [-h] binary ...
cargo-runner.py: error: the following arguments are required: binary, arguments
PS> cargo test
    Finished test [optimized + debuginfo] target(s) in 0.03s
     Running unittests src\lib.rs (target\thumbv7em-none-eabihf\debug\deps\flipperzero-b5d0b378c3208e7a)
error: test failed, to rerun pass `-p flipperzero --lib`

Caused by:
  could not execute process `C:\Users\dcoles\workspace\flipperzero-rs\crates\./cargo-runner.py C:\Users\dcoles\workspace\flipperzero-rs\crates\target\thumbv7em-none-eabihf\debug\deps\flipperzero-b5d0b378c3208e7a` (never executed)

Caused by:
  %1 is not a valid Win32 application. (os error 193)
```

In this case, maybe we just go with the shell script and make this work for Windows in a future PR.

crates/cargo-runner.py
str4d: Oops, that's left over from debugging.

str4d:
> > it should now also work on Windows (untested, but [Windows can execute *.py files directly if their extension is associated correctly](https://docs.python.org/3/faq/windows.html#how-do-i-make-python-scripts-executable)).
> 
> Looks like `runner` must be an application, not just an associated file type:

I've now found https://github.com/rust-lang/cargo/issues/8347 which indicates we just need to set the runner to `python path/to/script.py`. I'll update the PR to do that, and then you can retest on Windows.

str4d:
> @str4d It looks like the runner [might be causing problems for GitHub Actions](https://github.com/flipperzero-rs/flipperzero/actions/runs/4581112299/jobs/8090398886?pr=57). Can you take a look?

I think what's going on here is that CI was configured to run all tests, but nothing had tests enabled so it did nothing. Now that we enable tests, CI tries to run them, but because the `.cargo/config` sets the default target to be not the host, they fail. I think the solution is to just not run any tests in CI, as there are no useful tests to be run outside the Flipper Zero.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/bab2e58543afd8d32c10081eaf576c0e803a9731..e869e356a17de0b18d961b2f4b2e72f6da83e028) to add missing package to CI, and hopefully fix `cargo test` on Windows.

str4d:
Okay, confirmed that CI fails now because it can't find a Flipper Zero. I'll adjust CI to expect that failure, so we're still compiling everything and checking the runner.

dcoles:
> I've now found https://github.com/rust-lang/cargo/issues/8347 which indicates we just need to set the runner to python path/to/script.py. I'll update the PR to do that, and then you can retest on Windows.

Looks like I hit python/cpython#99185 (Windows does not provide a `python3.exe`), which is totally dumb and bites me all the time. It would be possible to make the script compatible with both Python 2 and 3, but on my Ubuntu box there is no `python` binary unless I install the `python-is-python3` package.

[PEP-397](https://peps.python.org/pep-0397) tries to resolve this situation by introducing a `py.exe` that can interpret the shbang line, but this isn't available on Ubuntu (not to be confused with [`pythonpy`](https://packages.ubuntu.com/jammy/pythonpy)).

Not sure if there's an easy fix. There doesn't seem to be a `cfg(host_os = "windows")` and I can't figure out a way to get `cmd.exe` and `bash` to accept the same `runner` line.

str4d:
Hmm. Now that I know we can have arguments inside `runner`, I wonder if we can just call `cargo` directly instead of using a wrapper script? The purpose of the script was to encapsulate the numerous arguments, but we also know that `cargo` should be available on the `PATH`. I'll give this a go.

str4d:
> The purpose of the script was to both encapsulate the numerous arguments and enable changing working directory so that the `.cargo/config` won't apply. If the latter can be worked around, then this might work.

I cannot figure out how to work around the latter inside `target.<target>.runner` without adding a `cd ... &&`, which is not cross-platform. However, reading a bunch of issues (around trying to unset `build.target` (https://github.com/rust-lang/cargo/issues/8112) or reset it to the host (https://github.com/rust-lang/cargo/issues/8687)) led me to discover [`per-package-target`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#per-package-target), which seems like it would also resolve the issue. We would set each of the crates under `crates/` to have a `package.forced-target`, and then remove the `build.target` from `crates/.cargo/config.toml`. This would mean that calling `cargo run --manifest-path ../tools/Cargo.toml` from inside `crates/` would then compile for the host instead of the Flipper Zero. However, it would block #58 on another unstable feature.

str4d:
Sadly `per-package-target` doesn't work, because even though `forced-target` correctly causes the test binary to be compiled for the Flipper Zero, it doesn't cause the corresponding per-target runner to be used (so instead the binary runs directly on the host). I've asked about this in the tracking issue (https://github.com/rust-lang/cargo/issues/9406#issuecomment-1493318519).

However, I've figured out another solution: aliases! We remove the `build.target` (so by default we build for the host), and then add aliases that set the correct `--target`. I've pushed a commit that does this, providing `*-flipper` aliases for `check`, `build`, and `test` (happy to rename to something else / shorter if preferred). This does interfere with muscle memory though (so it's a trade-off between "need to remember separate commands" with this commit, and "need to make custom changes to my OS environment to get Python working" before), and it means that e.g. the `rust-analyzer` plugin for VSCode needs to be configured to use the correct target (I can commit that if you want).

str4d:
@dcoles is there anything else I need to do before this can be merged? I'd love to get back on with my other PRs like #50 that are blocked on testing.

str4d:
> Over the weekend I got to see how a few other projects tackle this. Ferrus Systems has a tool called [`probe-run`](https://crates.io/crates/probe-run) that can be used as a runner for a number of embedded platforms (though probably not compatible with Flipper apps). You can use the [`hf2-cli`](https://crates.io/crates/hf2-cli) in a similar way via `hf2 elf`.

Yeah, those were what made me realize we could run tests and apps in this way. But we can't use any of the existing ones because they flash firmware, and I wanted this to be as easy to use as possible, so no need to install anything ideally. Didn't quite get there on all platforms, and if we end up having an installable FAP runner then I don't mind.

str4d:
The other thing is that an installable FAP runner will be much nicer once upstream has implement proper RPC support for FAPs, so we don't have to do the hacky "write stdout to a file and then copy it off afterwards" that really is only suitable for one-shot FAPs.

dcoles:
> The other thing is that an installable FAP runner will be much nicer once upstream has implement proper RPC support for FAPs, so we don't have to do the hacky "write stdout to a file and then copy it off afterwards" that really is only suitable for one-shot FAPs.

I'm very much looking forward to that. It's bit awful what you have to do when trying to use a serial CLI as a file transfer layer.

PR #60, opened 2023-04-26
More wrappers

str4d:


crates/flipperzero/src/furi/rng.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// A random number generator that retrieves randomness from the Flipper Zero hardware.
+///
+/// This is a zero-sized struct. It can be freely constructed with `HwRng`.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct HwRng;
+
+impl CryptoRng for HwRng {}
+
+impl RngCore for HwRng {
```
str4d, line 12: Instead of (or in addition to) this approach, we could provide a standlone function that matches the API of `getrandom::getrandom`, and then the logic that sets up the root-level FAP could load that in as a [custom `getrandom` backend](https://docs.rs/getrandom/0.2.9/getrandom/macro.register_custom_getrandom.html). This would enable any usages of `getrandom::getrandom` to work, as well as `rand_core::OsRng` / `rand::rngs::OsRng`.

crates/flipperzero/src/toolbox/md5.rs
```diff
@@ -0,0 +1,126 @@
...
+        unsafe { sys::md5_starts(&mut state) };
+        Self { state }
+    }
+}
+
+impl UpdateCore for Md5Core {
+    #[inline]
+    fn update_blocks(&mut self, blocks: &[Block<Self>]) {
+        for block in blocks {
+            self.state.total[0] += Self::BlockSize::U32;
```
dcoles, line 71: I'm honestly a bit lost by this syntax (the docs are pretty opaque). `Self::BlockSize` is `typenum::U64`, so that makes `Self::BlockSize::U32`  equivalent to `typenum::U64::U32`. Is this just a compile-time type representing `64u32`?

If so, is `Self::BlockSize::to_u32()` equivalent (though an inline function vs. constant)?

crates/flipperzero/src/toolbox/sha256.rs
```diff
@@ -0,0 +1,118 @@
...
+            self.state.total[0] += Self::BlockSize::U32;
+            if self.state.total[0] < Self::BlockSize::U32 {
+                self.state.total[1] += 1;
+            }
+
+            unsafe {
+                sys::memcpy(
+                    self.state.wbuf.as_mut_ptr().cast(),
+                    block.as_ptr().cast(),
+                    Self::BlockSize::U32,
```
dcoles, line 64: I'm assuming this evaluates to `64u32` Would `Self::BlockSize::USIZE` match the signature better here?

dcoles (approved):
Looks great!

I needed a few cups of coffee to get my head around the `typenum` use, but it looks right to me.

crates/flipperzero/src/toolbox/md5.rs
str4d: > Is this just a compile-time type representing `64u32`?

Yes. Rust lacked sufficient const generics to perform arithmetic in the type system using integers (and even for very recent Rust versions it still lacks a few things that some use cases need). `typenum` is a workaround that uses types instead,.

> If so, is `Self::BlockSize::to_u32()` equivalent (though an inline function vs. constant)?

Yep. The associated constant is provided for use in `const fn`s, and I use it because it's shorter, but I can use the function instead if you prefer (which adds a bit of indirection, but the compiler _should_ inline it).

crates/flipperzero/src/toolbox/sha256.rs
str4d: If we used [`core::ptr::copy_nonoverlapping`](https://doc.rust-lang.org/core/ptr/fn.copy_nonoverlapping.html) instead then yes. But `sys::memcpy` takes [`core::ffi::c_uint`](https://doc.rust-lang.org/stable/core/ffi/type.c_uint.html) as its third argument, which is a `u32`.

dcoles: Makes sense. I have to remember to check the Flipper APIs vs. the Linux ones.

crates/flipperzero/src/toolbox/md5.rs
dcoles: Nice. Thanks for explaining. I think it's fine to keep it the current way.

Maybe throw in a `// 64u32` next to the line just to nudge people in the right direction.

crates/flipperzero/src/toolbox/sha256.rs
```diff
@@ -0,0 +1,118 @@
...
+                );
```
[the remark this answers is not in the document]
dcoles: It looks like regardless of using `sys::memcpy` or `core::ptr::copy_nonoverlapping` the end result is same. They both pull in about 648 bytes of `compiler_builtins` implementation:

```
000000e8 <__aeabi_memcpy>:
  e8:   f7ff bffe       b.w     ec <compiler_builtins::arm::__aeabi_memcpy::h79aa9afdd08ccabc>

000000ec <compiler_builtins::arm::__aeabi_memcpy::h79aa9afdd08ccabc>:
  ec:   f7ff bffe       b.w     f0 <compiler_builtins::mem::memcpy::h52732c43e1a322b3>

000000f0 <compiler_builtins::mem::memcpy::h52732c43e1a322b3>:
  f0:   b5f0            push    {r4, r5, r6, r7, lr}
  f2:   af03            add     r7, sp, #12
  ...
```

So it looks like it doesn't really matter one way or the other.

PR #61, opened 2023-05-04
tools/storage: Create directory before sending file

mogenson:
The `build-examples.py --install` script would fail on my Flipper with: "ERROR: file/dir not exist" because there was no Examples directory.

Make a small change to storage.rs to get the parent path on the Flipper SD card for the file to send, and create that directory before sending the file.

Converting between FlipperPaths and Paths or PathBufs is a pain, so just split from the last "/" character to get the parent path.

### DIFF
--- a/tools/src/storage.rs
+++ b/tools/src/storage.rs
@@ -2,9 +2,9 @@
 
 use std::fmt::Display;
 use std::io::{Read, Write};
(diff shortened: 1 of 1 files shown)


dcoles:
Thanks for the PR! Much appreciated.

PR #62, opened 2023-05-07
feat: support for API 23.0

ssoudan:
- Dockerfile to build the binding from the git
- added instruction to update  bindings.rs in README.md
- `bindings.rs` updated to API 23.0


README.md
```diff
@@ -1,7 +1,7 @@
 # Rust for Flipper Zero 🐬❤️🦀
 
 [![crates.io](https://img.shields.io/crates/v/flipperzero)](https://crates.io/crates/flipperzero)
-[![Flipper Zero API](https://img.shields.io/badge/Flipper%20Zero%20API-20.0-orange)](https://github.com/flipperdevices/flipperzero-firmware/blob/0.80.1/firmware/targets/f7/api_symbols.csv)
+[![Flipper Zero API](https://img.shields.io/badge/Flipper%20Zero%20API-26.0-orange)](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/firmware/targets/f7/api_symbols.csv)
```
str4d, line 4: Pinning the URL to `dev` means the `api_symbols.csv` won't be stable relative to the symbols used by this repo's `main` branch. This should be bumped to a firmware revision that has the API 26.0 symbols. Given that it appears the current stable firmware is only on API 23, I suggest we only bump to there for now.

crates/flipperzero/src/io.rs
```diff
@@ -39,6 +39,27 @@ pub enum Error {
     Uncategorized(sys::FS_Error),
 }
 
+impl ufmt::uDisplay for Error {
```
str4d, line 42: The PR needs rebasing, and that will move it on top of 9ae08559ef2a7060cde78da7efb04e8ce758c91fwhich includes this impl.

README.md
```diff
@@ -17,19 +17,26 @@ This means it's not possible to use anything in the [`std`](https://doc.rust-lan
...
 | 0.5.x         | 10.1        |
 | 0.4.x         | 7.5         |
 | 0.3.x         | 2.2         |
 
+To build a newer version of the bindings:
+
+```shell
+docker run --rm -t $(docker build -q .) > crates/sys/src/bindings.rs
+```
+
```
str4d, line 39: Instead of putting this content here, please amend `docs/updating-sdk.md` to describe the Docker flow alongside the non-Docker flow.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    clang \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+RUN git clone --recursive $FIRMWARE_GIT firmware
+
+WORKDIR /app/firmware
+
+RUN ./fbt
```
str4d, line 21: This is going to build whatever happens to be in the default branch (the `dev` branch, which in particular has no guarantees of stability). It should instead take a specific tag corresponding to the firmware we want the bindings to be API-compatible with.

.dockerignore
```diff
@@ -0,0 +1 @@
+Dockerfile
```
str4d, line 1: I'm not sure we want the top-level `Dockerfile` to be the bindings generator; I can imagine other Docker workflows that we might end up wanting in the repo. Could this be moved to a sub-directory?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+RUN cp bindings.rs ../crates/sys/src
+
+WORKDIR /app/flipperzero-rs/crates
+
+# test it builds
+RUN cargo build --release
+
+####################
+ENTRYPOINT [ "cat", "sys/src/bindings.rs" ]
```
str4d, line 56: The generated bindings having no formatting whatsoever, which means that the usage steps are incomplete and multi-step. Is the problem that the generated bindings inside Docker are not formatted (in which case the Dockerfile needs a `cargo fmt` step), or is it a problem with this method of extracting them?

str4d (changes requested):

README.md
ssoudan: Will use 0.82.3 tag instead.

README.md
ssoudan: make sense

crates/flipperzero/src/io.rs
ssoudan: on it

Dockerfile
ssoudan: adding another ARG to the Dockerfile.

Dockerfile
ssoudan: will get it formatted in the Dockerfile.

.dockerignore
ssoudan: Moved to `tools/` updated `tools/README.md` and the command in `docs/updating-sdk.md`.

crates/flipperzero/src/io.rs
ssoudan: rebased

examples/hello-storage/Cargo.lock
str4d, line 1: This file was deleted in ccc4a4a529eb5bd4699a94c974731b543007dd34; I presume it was mistakenly re-added during the rebase.

examples/hello-storage/rust-toolchain.toml
str4d, line 1: This file was deleted in ccc4a4a529eb5bd4699a94c974731b543007dd34; I presume it was mistakenly re-added during the rebase.

crates/sys/src/bindings.rs
str4d, line 1: Every single line of this file is marked with a change, and `git` shows every line has a `^M` added to the end (indicating the line endings are `CRLF` instead of `LF`. Something about the workflow causes them to be added, and we do not want them in the committed file. Did you run the Dockerfile from Windows?

docs/updating-sdk.md
```diff
@@ -29,8 +29,15 @@ $ cargo run --bin generate-bindings ../../flipperzero-firmware/build/f7-firmware
...
-Make sure to update the SDK details in [`README.md`] before making a new release.
+Alternatively, you can generate `binding.rs` in an isolated env using Docker and the following command:
+
+From the root of the repository, to build the binding for the branch/tag `0.82.3` of the official SDK:
 
+```shell
+docker run --rm -t $(docker build --build-arg BRANCH=0.82.3 -q -f tools/Dockerfile .) > crates/sys/src/bindings.rs
+```
+
+Make sure to update the SDK details in [`README.md`] before making a new release.
```
str4d, line 40: I also don't understand why Git marks this line as having changed. Maybe another CRLF problem? How is your local `git` configured?

str4d (changes requested):

examples/hello-storage/Cargo.lock
ssoudan: indeed

examples/hello-storage/Cargo.toml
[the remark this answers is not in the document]
ssoudan: same

crates/sys/src/bindings.rs
ssoudan: added `--config newline_style=unix` to `rustfmt` in the `Dockerfile`. And reformatted the file.

docs/updating-sdk.md
ssoudan: Changed the order of the lines.

str4d (approved):
utACK (but I did generate the bindings for API 23.0 locally and confirmed they matched what this PR shows).

ssoudan:
Thanks!

PR #63, opened 2023-05-08
Updates to enable publishing test crates

str4d:
`flipperzero-test 0.1.0` and `flipperzero-test-macros 0.1.0` have been published from d3f48ebeeb54ffac442740998bd1575c83ed819e. This should enable downstream users to write their own tests (maybe already, maybe with some future improvement PRs).

### DIFF
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "sys",
     "rt",
     "test",
+    "test/macros",
 ]
 resolver = "2"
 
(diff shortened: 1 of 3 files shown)


str4d:
@dcoles I've sent you an owner invite on crates.io for both of these crates (I mainly wanted these published to ensure the crate names were obtained, but also because I eventually want to use these in my own downstream libraries). I published as 0.1.0 for now, but we can synchronise the versions if you prefer.

dcoles:
@str4d Thanks! I've added the @flipperzero-rs/maintainers group as an owner for this and the other flipperzero-rs crates. That should have granted you ownership to those packages too.

PR #65, opened 2023-05-10
Release 0.9.0

str4d:


str4d:
I'm keen to get #34 into this release, but given that it's purely additive, it could also go into a 0.9.1.

dcoles:
I really need to hurry up and get the bindings building automatically. I now have it fetching the SDK/toolchain, I just have to update all the paths so it works at build time.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/66fa71b2f24e47d015399584119a8d29fcfc75b7..a4831075812f9a7ac7fbacc1089c4da32cd86998) to include #34, and finished filling in the changelog.

str4d:
@dcoles I _think_ I've done everything necessary for a release (based on looking at previous release commits).

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/7b7877cd51be26f1f76d2f6b7eef3245959a326c..b838a42a2760956d5152b8dd7364fc827ed56099) to bump MSRV to 1.70, which I realised we can do even though it hasn't been released to stable because we still require nightly.

dcoles:
> @dcoles I _think_ I've done everything necessary for a release (based on looking at previous release commits).

I get a small diff when I generate the bindings myself, but I'm a bit confused ([`api_symbols.csv` says both arguments are `size_t`](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@0.82.3/-/blob/firmware/targets/f7/api_symbols.csv?L2565:12)). Though `strlen` is also `core::ffi::c_uint`.

```diff
diff --git a/crates/sys/src/bindings.rs b/crates/sys/src/bindings.rs
index 227b529..8885697 100644
--- a/crates/sys/src/bindings.rs
+++ b/crates/sys/src/bindings.rs
@@ -1936,8 +1936,8 @@ extern "C" {
     pub fn strlcpy(
         arg1: *mut core::ffi::c_char,
         arg2: *const core::ffi::c_char,
-        arg3: usize,
-    ) -> usize;
+        arg3: core::ffi::c_uint,
+    ) -> core::ffi::c_uint;
 }
 extern "C" {
     #[doc = "Get free heap size\n\nReturns:\n\n* free heap size in bytes\n\n"] 
```

What version of the toolchain are you using? When I `git reset --hard 0.82.3`, remove `toolchain` and run `fbt` it downloads version 21:

```ps1
cat .\toolchain\x86_64-windows\VERSION
21
```

str4d:
AFAIK the bindings were generated using the Dockerfile introduced in #62, which uses Debian Bookworm as a base. I just tried regenerating them via the old method on my local machine (Ubuntu 22.04) and I get identical bindings.
```
$ git show HEAD
commit b7f406d43beb1beedc71282228ea148b84d16457 (HEAD -> release, tag: 0.82.3, origin/release)
Merge: 376319cb 849aa0f7
Author: Aleksandr Kutuzov <alleteam@gmail.com>
Date:   Fri Apr 28 00:46:14 2023 +0900

    Merge branch 'release-candidate' into release


$ cat ./toolchain/x86_64-linux/VERSION 
21
```

This sounds like a Linux-Windows divergence.

str4d:
I confirmed that the toolchain `./fbt` fetched for me uses `size_t` for both of those fields:
```
$ rg strlcpy toolchain/
toolchain/x86_64-linux/libtool/share/libtool/ltdl.c
1243:      strlcpy (dir, canonical, dirlen);

toolchain/x86_64-linux/libtool/share/aclocal/ltdl.m4
408:AC_CHECK_FUNCS([strlcat strlcpy], [], [AC_LIBOBJ([lt__strl])])

toolchain/x86_64-linux/libtool/share/libtool/loaders/dyld.c
353:      strlcpy (saveError, dylderror (LT__STRERROR (SYMBOL_NOT_FOUND)), 255);

toolchain/x86_64-linux/libtool/share/libtool/libltdl/lt__strl.h
49:#  define strlcpy(dst,src,dstsize) lt_strlcpy(dst,src,dstsize)
50:LT_SCOPE size_t lt_strlcpy(char *dst, const char *src, const size_t dstsize);

toolchain/x86_64-linux/libtool/share/libtool/lt__dirent.c
55:  if (lt_strlcpy (file_spec, path, sizeof file_spec) >= sizeof file_spec
98:  if (lt_strlcpy (entry->file_info.d_name, entry->Win32FindData.cFileName,

toolchain/x86_64-linux/libtool/share/libtool/lt__strl.c
85:  lt_strlcpy copies up to dstsize - 1 characters from the NULL-terminated
89:  substitutes for strlcpy(), which is available under OpenBSD, FreeBSD
94:    if (lt_strlcpy(dst, src, dstsize) >= dstsize)
99:lt_strlcpy(char *dst, const char *src, const size_t dstsize)

toolchain/x86_64-linux/libtool/share/libtool/config-h.in
100:/* Define to 1 if you have the `strlcpy' function. */

toolchain/x86_64-linux/libtool/share/libtool/configure
13179:for ac_func in strlcat strlcpy

toolchain/x86_64-linux/openssl/include/openssl/buffer.h
34:# define BUF_strlcpy(dst, src, size)  OPENSSL_strlcpy(dst, src, size)

toolchain/x86_64-linux/openssl/include/openssl/crypto.h
146:size_t OPENSSL_strlcpy(char *dst, const char *src, size_t siz);

toolchain/x86_64-linux/python/include/python3.11/pyconfig.h
1192:/* Define to 1 if you have the `strlcpy' function. */

toolchain/x86_64-linux/arm-none-eabi/include/string.h
116:size_t      strlcpy (char *, const char *, size_t);

str4d:
And if you look at the API update commit, you can see the exact opposite diff occurring, which implies it is a difference across API versions:
https://github.com/flipperzero-rs/flipperzero/commit/f4c818c7627bf1ec5b96ddae9ec90e958364717e#diff-babe1ed57a4d43e72cade256fcf1d86d5bb986931aaee7e372fa327af7e49018L1939-R1940

dcoles:
> This sounds like a Linux-Windows divergence.

I think I agree. I forced `generate-bindings.rs` to use `toolchain/x86_64-linux/arm-none-eabi/include` and then ran `generate-bindings` both Windows native and on the same folder using WSL2. The WSL2 version (Linux) matches your output.

Same version of `bindgen` and `clang-sys`, so possibly some weird Windows/Linux path resolution thing with includes?

Anyway, I'd rather go with the Linux version than the Windows one. 👍

dcoles (approved):

dcoles:
@str4d Updated my Flipper Zero to 0.82.3 and ran the test suite—all good!

Though it looks like some of the example apps (e.g. GUI example) are hanging at exit (stuck on the ⏳ spinner).
We should at least determine the cause of this before pushing 0.9.0 to crates.io.

dcoles:
So I added some logging and it appears that it's blocked on the `GuiSrv` thread.

I suspect I know the issue here. The GUI service's [appid is `gui`] and our GUI example also has the appid `gui`.
(This also explains why the notification example hangs too).

dcoles:
> Aha, my assumption that they'd domain-separate FAPs from their system threads was wrong.
> 
> Well, there's nothing we can do about that inside the crates (we can't change the main thread's app ID, and we don't currently have heap statics for collecting the app IDs of the threads we might create). So maybe just document this as a hazard for now?

It's a little crude, but one option would be to ignore any thread whose name ends in `"Srv"`. Otherwise it's a little too easy to have a conflict.

Being able to check if a thread is a service is one reason it might make sense to upstream this logic. Though they might also say that there's no guarantee that `appid` is ever unique.

dcoles:
I've added the workaround to ignore threads ending in "Srv" 9bcd38399438e4ebf8e4b50f09962341a1ce99ba and now all the examples exit successfully. While I do think hanging at exit is more obvious than just crashing, letting threads outlive the lifetime of the main thread is still a bug in the app. Short of changing the thread API to something like [`std::thread::scope`](https://doc.rust-lang.org/stable/std/thread/fn.scope.html), I don't know if we can 100% ensure that the user uses threads in the correct manner.

Crates are now published as [flipperzero-0.9.0](https://crates.io/crates/flipperzero)! 🥳

PR #66, opened 2023-05-13
Add format! macro

kitterion:
Now that `FuriString` exists, a convenience `format!` macro should also exist.


dcoles (approved):

dcoles:
This looks great! Thank you for the MR!

PR #67, opened 2023-05-13
Use ptr::NonNull instead of raw pointers

kitterion:
This allows rust's null pointer optimization to kick in. For example, with this change the size of `Option<FuriString>` becomes 8 -> 4.
As far as I can tell, flipperzero's malloc doesn't ever return null but crashes the app instead.
See https://github.com/flipperdevices/flipperzero-firmware/blob/0.82.3/furi/core/memmgr_heap.c#L484
That's the function that `malloc` calls directly, it has the only return statement at the very end, and `furi_check` just crashes the app if the condition is false, i.e. the pointer is null.

I think I caught all the structs, I'm happy to add more if I didn't.

### DIFF
--- a/crates/flipperzero/src/dialogs/mod.rs
+++ b/crates/flipperzero/src/dialogs/mod.rs
@@ -5,7 +5,7 @@ use alloc::ffi::CString;
 
 use core::ffi::{c_char, CStr};
 use core::marker::PhantomData;
-use core::ptr;
+use core::ptr::{self, NonNull};
 
 use flipperzero_sys as sys;
 use flipperzero_sys::furi::UnsafeRecord;
@@ -24,7 +24,7 @@ pub struct DialogsApp {
 
 /// A dialog message.
 pub struct DialogMessage<'a> {
-    data: *mut sys::DialogMessage,
+    data: NonNull<sys::DialogMessage>,
     _phantom: PhantomData<&'a CStr>,
 }
 
@@ -46,7 +46,8 @@ impl DialogsApp {
 
     /// Displays a message.
     pub fn show(&mut self, message: &DialogMessage) -> DialogMessageButton {
-        let button_sys = unsafe { sys::dialog_message_show(self.data.as_ptr(), message.data) };
+        let button_sys =
+            unsafe { sys::dialog_message_show(self.data.as_ptr(), message.data.as_ptr()) };
 
         DialogMessageButton::from_sys(button_sys).expect("Invalid button")
     }
@@ -55,8 +56,7 @@ impl DialogsApp {
 impl<'a> DialogMessage<'a> {
     /// Allocates a new dialog message.
     pub fn new() -> Self {
-        let data = unsafe { sys::dialog_message_alloc() };
-        assert!(!data.is_null());
+        let data = unsafe { NonNull::new_unchecked(sys::dialog_message_alloc()) };
 
         Self {
             data,
@@ -76,7 +76,7 @@ impl<'a> DialogMessage<'a> {
         let right = right.map_or(ptr::null(), |l| l.as_ptr());
 
         unsafe {
-            sys::dialog_message_set_buttons(self.data, left, center, right);
+            sys::dialog_message_set_buttons(self.data.as_ptr(), left, center, right);
         }
     }
 
@@ -91,7 +91,7 @@ impl<'a> DialogMessage<'a> {
     ) {
         unsafe {
             sys::dialog_message_set_header(
-                self.data,
+                self.data.as_ptr(),
                 header.as_ptr(),
                 x,
                 y,
@@ -105,7 +105,7 @@ impl<'a> DialogMessage<'a> {
     pub fn set_text(&mut self, text: &'a CStr, x: u8, y: u8, horizontal: Align, vertical: Align) {
         unsafe {
             sys::dialog_message_set_text(
-                self.data,
+                self.data.as_ptr(),
                 text.as_ptr(),
                 x,
                 y,
@@ -119,7 +119,7 @@ impl<'a> DialogMessage<'a> {
     pub fn clear_header(&mut self) {
         unsafe {
             sys::dialog_message_set_header(
-                self.data,
+                self.data.as_ptr(),
                 ptr::null(),
                 0,
                 0,
@@ -133,7 +133,7 @@ impl<'a> DialogMessage<'a> {
     pub fn clear_text(&mut self) {
         unsafe {
             sys::dialog_message_set_text(
-                self.data,
+                self.data.as_ptr(),
                 ptr::null(),
                 0,
                 0,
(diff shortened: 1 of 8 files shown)


dcoles:
Thanks for the PR. Yes, from what I've seen `malloc` will never fail and the memory is always zeroed regardless of whether you use `malloc` or `calloc`.

crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -17,7 +18,10 @@ impl<M: Sized> MessageQueue<M> {
     pub fn new(capacity: usize) -> Self {
         Self {
             hnd: unsafe {
-                sys::furi_message_queue_alloc(capacity as u32, core::mem::size_of::<M>() as u32)
+                NonNull::new_unchecked(sys::furi_message_queue_alloc(
```
str4d, line 21: This calls `xQueueCreate` and then uses `furi_check` to enforce non-null.

crates/flipperzero/src/furi/string.rs
```diff
@@ -57,7 +57,7 @@ impl FuriString {
     #[inline]
     #[must_use]
     pub fn new() -> Self {
-        FuriString(unsafe { sys::furi_string_alloc() })
+        FuriString(unsafe { NonNull::new_unchecked(sys::furi_string_alloc()) })
```
str4d, line 60: This calls `malloc`.

crates/flipperzero/src/furi/string.rs
```diff
@@ -227,7 +227,9 @@ impl FuriString {
         // SAFETY: Trimming the beginning of a C string results in a valid C string, as
         // long as the nul byte is not trimmed.
         assert!(at <= self.len());
-        let ret = FuriString(unsafe { sys::furi_string_alloc_set_str(self.as_c_ptr().add(at)) });
+        let ret = FuriString(unsafe {
+            NonNull::new_unchecked(sys::furi_string_alloc_set_str(self.as_c_ptr().add(at)))
```
str4d, line 231: This calls `malloc`.

crates/flipperzero/src/furi/string.rs
```diff
@@ -485,7 +487,7 @@ impl FuriString {
 
 impl Clone for FuriString {
     fn clone(&self) -> Self {
-        Self(unsafe { sys::furi_string_alloc_set(self.0) })
+        Self(unsafe { NonNull::new_unchecked(sys::furi_string_alloc_set(self.0.as_ptr())) })
```
str4d, line 490: This calls `malloc`.

crates/flipperzero/src/furi/sync.rs
```diff
@@ -17,16 +18,13 @@ type UnsendUnsync = PhantomData<*const ()>;
...
     data: UnsafeCell<T>,
 }
 
 impl<T> Mutex<T> {
     pub fn new(data: T) -> Self {
-        let mutex = unsafe { sys::furi_mutex_alloc(MUTEX_TYPE) };
-        if mutex.is_null() {
-            panic!("furi_mutex_alloc failed");
-        }
+        let mutex = unsafe { NonNull::new_unchecked(sys::furi_mutex_alloc(MUTEX_TYPE)) };
```
str4d, line 27: This calls either `xSemaphoreCreateMutex` or `xSemaphoreCreateRecursiveMutex`, and then uses `furi_check` to enforce non-null.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -149,7 +154,7 @@ where
 pub fn current() -> Thread {
     use alloc::borrow::ToOwned;
 
-    let thread = unsafe { sys::furi_thread_get_current() };
+    let thread = unsafe { NonNull::new_unchecked(sys::furi_thread_get_current()) };
```
str4d, line 157: This calls `pvTaskGetThreadLocalStoragePointer(NULL, 0)`. It does not check that the result is non-null, and there are other places inside the Flipper Zero firmware that call `pvTaskGetThreadLocalStoragePointer` and behave conditionally if it is null. However, the thread pointer is set via `vTaskSetThreadLocalStoragePointer` as the first action in `furi_thread_body` before calling `thread->callback`, and it should never be possible to call `thread::current()` in a thread outside of `thread->callback`, so this should be sound.

crates/flipperzero/src/furi/thread.rs
```diff
@@ -218,7 +223,10 @@ impl Thread {
                     sys::furi_thread_enable_heap_trace(thread);
                 }
             }
-            Thread { name, thread }
+            Thread {
+                name,
+                thread: NonNull::new_unchecked(thread),
```
str4d, line 228: `thread` is obtained from `furi_thread_alloc`, which calls `malloc`.

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -55,8 +56,7 @@ impl DialogsApp {
 impl<'a> DialogMessage<'a> {
     /// Allocates a new dialog message.
     pub fn new() -> Self {
-        let data = unsafe { sys::dialog_message_alloc() };
-        assert!(!data.is_null());
+        let data = unsafe { NonNull::new_unchecked(sys::dialog_message_alloc()) };
```
str4d, line 59: This calls `malloc`.

crates/flipperzero/src/storage.rs
```diff
@@ -141,19 +142,22 @@ impl OpenOptions {
...
-pub struct File(*mut sys::File, UnsafeRecord<sys::Storage>);
+pub struct File(NonNull<sys::File>, UnsafeRecord<sys::Storage>);
 
 impl File {
     pub fn new() -> Self {
         unsafe {
             let record = UnsafeRecord::open(RECORD_STORAGE);
-            File(sys::storage_file_alloc(record.as_ptr()), record)
+            File(
+                NonNull::new_unchecked(sys::storage_file_alloc(record.as_ptr())),
```
str4d, line 158: This calls `malloc`.

str4d (commented):
Post-hoc utACK.

dcoles:
> Post-hoc utACK.

I'm not familiar with utACK. Is that "untested approval"?

I also took a look through each of the functions to verify that it either calls `furi_check` to ensure the values are non-null or called `malloc` ([which internally calls `furi_check` as well as zeros out the allocated memory](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@341cd5e860a67b70b46a87d79d705caa1d457155/-/blob/furi/core/memmgr_heap.c?L484-485))

I think `NonNull` is the right choice for these structures, since they can't possibly work with a null value (and the `furi_check`s allow us to bypass the need for using the checked [`NonNull::New`](https://doc.rust-lang.org/core/ptr/struct.NonNull.html#method.new)).

(I think you are agreeing, it's just not clear given the unresolved comments / unapproved review.)

str4d:
> I'm not familiar with utACK. Is that "untested approval"?

Yep. Just an indication that it was a visual review only, and I did not e.g. run the code myself to check how it behaves.

> I think you are agreeing,

Yes, I am. I just wanted to check myself that the changes made sense (because `NonNull` has more stringent requirements), and leave a record of why I was satisfied.

> it's just not clear given the unresolved comments / unapproved review.

After a PR is merged, GitHub doesn't let anyone submit either "Approved" or "Changes Requested" PR reviews, so the above was submitted as a "Comments" review. But it would have been "Approved", hence (ut)ACK.

PR #68, opened 2023-05-16
serial_cli: capture exit keycode on MacOS terminal

mogenson:
On MacOS with the default Terminal application, Crossterm reports the
`Ctrl + ]` sequence as `Ctrl + 5`. The `fbt` Python tool seems to work
fine on MacOS + Terminal app, so this may be a Crossterm quirk.
    
Capture the `Ctrl + 5` keycode and exit process.

### DIFF
--- a/tools/src/bin/serial_cli.rs
+++ b/tools/src/bin/serial_cli.rs
@@ -81,7 +81,8 @@ fn run(port: &mut dyn serialport::SerialPort) -> io::Result<()> {
                     }
 
                     match (modifiers, code) {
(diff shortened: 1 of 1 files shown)


dcoles (approved):

dcoles:
Hi @mogenson. Thanks for the PR!

I honestly not an expert on keyboard mappings, but it seems that in many (most?) terminal keyboard mappings (e.g. US) `Ctrl+]` and `Ctrl+5` both map to `Control_bracketright`. So it seems like it should be fine to accept either.

PR #69, opened 2023-05-17
Add ViewDispatcher example

mogenson:
Add an example demonstrating a simple use of the ViewDispatcher module. This app switches between two views. A TextInput view that prompts the user to enter their name, and a Widget view that says hello to the name.

### DIFF
--- a/crates/flipperzero/examples/view_dispatcher.rs
+++ b/crates/flipperzero/examples/view_dispatcher.rs
@@ -1,0 +1,134 @@
+//! Demonstrates use of the ViewDispatcher module.
+//!
+//! This app prompts the user for a name then says hello.
+
+#![no_main]
+#![no_std]
+
+extern crate alloc;
+extern crate flipperzero_alloc;
+extern crate flipperzero_rt;
+
+use alloc::boxed::Box;
+use core::ffi::{c_char, c_void, CStr};
+use core::ptr::NonNull;
+use flipperzero::furi::string::FuriString;
+use flipperzero_rt::{entry, manifest};
+use flipperzero_sys as sys;
+
+const RECORD_GUI: *const c_char = sys::c_string!("gui");
+
+manifest!(name = "Rust ViewDispatcher example");
+entry!(main);
+
+enum AppView {
+    Widget = 0,
+    TextInput = 1,
+}
+
+struct App {
+    name: [c_char; 16],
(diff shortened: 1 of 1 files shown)


crates/flipperzero/examples/view_dispatcher.rs
```diff
@@ -0,0 +1,130 @@
...
+manifest!(name = "Rust ViewDispatcher example");
+entry!(main);
+
+enum AppView {
+    Widget = 0,
+    TextInput = 1,
+}
+
+struct App {
+    name: [i8; 16],
```
dcoles, line 29: You might want to use [`core::ffi::c_char`](https://doc.rust-lang.org/stable/core/ffi/type.c_char.html) here if this is a C-string.

dcoles (commented):
@mogenson Thanks for the PR!

Just had a couple of minor points, otherwise all looks great.

crates/flipperzero/examples/view_dispatcher.rs
mogenson: This is a buffer provided to the `TextInput` to populate with a null-terminated C-string. I'm using a Rust array so that Rust allocates it as part of the Box and takes care of the lifetime.

crates/flipperzero/examples/view_dispatcher.rs
```diff
@@ -0,0 +1,130 @@
...
+    view_dispatcher: *mut sys::ViewDispatcher,
+    widget: *mut sys::Widget,
+    text_input: *mut sys::TextInput,
```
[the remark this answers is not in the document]
mogenson: Thanks for the suggestion! I'm trying to learn some embedded Rust so it's great to see useful features and strategies.

crates/flipperzero/examples/view_dispatcher.rs
dcoles: Ah. `core::ffi::c_char` should be the same as `i8` (usually). So `[c_char; 16]` should behave exactly the same.

It just helps make it clear that this array is being used as an array of `char` rather than an array of numbers.

mogenson: Ok, I understand now. I'll change the buffer type to `[c_char; 16]`.

dcoles:
Thanks for the PR!

Issue #41, opened 2023-02-10
build: automate bindings generation


JarvisCraft:
# Description

## Introduction

Bindings are currently generated using `tools/` based on bindgen. This works okay but have the following cons:

- any updates to original flipper headers require manual regeneration of the sources:
  - this also requires some maintainer (i.e. @dcoles) aware of the right generation process (including correct file and environment setup);
- this is hardly reproducible due to non-trivial build order;
- this process is not transparent.

## The suggestion

The idea is to replace this approach with a `build.rs`-based generation.

I expect this to be in a form of a cross-platform `build.rs` in `crates/sys` capable of running codegeneration by itself.

### Original definitions location

It's hard for me to decide what is the best way to store files with original header-definitions. I guess that it is worth looking on other projects utilizing `build.rs` for FFI code generation.

One way may be using git submodules, although I look at this as not the best mechanism which is hard to maintain sometimes so looking at alternative approaches is something expected.

We may also be able to utilize [`api_symbols.csv`](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/firmware/targets/f7/api_symbols.csv) provided in firmware repository as it seems to list available functions, but it seems to lack any `struct` definitions and type aliases which is a serious limitation.

### Version updates

It is also worth discussing the mechanism for performing version updates.

I would suggest following dependabot's approach of performing scheduled update checks. We could theoretically configure GitHub Actions to run on schedule checking for fresh firmware releases and, in case of them being present, run automation to create a PR with header updates.

dcoles:
Hi @JarvisCraft,

I took a bit of a look into how a package like [`openssl-sys`](https://crates.io/crates/openssl-sys) handle this problem.

In short, they build against the system OpenSSL headers and [provide no guarantee of any API stability when linking against one specific version or another](https://docs.rs/openssl/latest/openssl/#feature-detection). This puts the burden on the higher level crates to bridge this gap.

They expose the currently linked version via a `DEP_OPENSSL_VERSION_NUMBER` build environment variable so that other crates can conditionally compile code depending on the version. Additionally there is also a `vendored` feature that uses the [`openssl-src`](https://crates.io/crates/openssl-src) crate to build OpenSSL, however it still requires someone to manually update a Git sub-module for each new version.

[...]

PR #70, opened 2023-05-17
Update bindings for firmware version 0.83.1

kitterion:
https://github.com/flipperdevices/flipperzero-firmware/releases/tag/0.83.1

From what I can tell, it's a relatively large diff because the order of a lot of structs and functions got changed without any actual changes. Seems like the order of include paths got shuffled around a bit probably because of this item in the release notes:
- Libs: STM32CubeWB replaced with it's stand-lone, stripped version

### DIFF
--- a/crates/sys/src/bindings.rs
+++ b/crates/sys/src/bindings.rs
@@ -110,7 +110,7 @@ impl<T> ::core::fmt::Debug for __IncompleteArrayField<T> {
         fmt.write_str("__IncompleteArrayField")
     }
 }
-pub const API_VERSION: u32 = 1507328;
+pub const API_VERSION: u32 = 1703939;
 pub type wint_t = core::ffi::c_int;
 pub type _off_t = core::ffi::c_long;
 pub type _fpos_t = core::ffi::c_long;
@@ -1778,7 +1778,8 @@ extern "C" {
     #[doc = "Delay in microseconds\nImplemented using Cortex DWT counter. Blocking and non aliased.\n\n# Arguments\n\n* `microseconds` - [Direction: In] microseconds to wait\n\n"]
     pub fn furi_delay_us(microseconds: u32);
 }
-pub type va_list = u32;
+pub type __gnuc_va_list = u32;
+pub type va_list = __gnuc_va_list;
 pub type FILE = __FILE;
 extern "C" {
     pub fn sscanf(
@@ -1936,8 +1937,8 @@ extern "C" {
     pub fn strlcpy(
         arg1: *mut core::ffi::c_char,
         arg2: *const core::ffi::c_char,
-        arg3: usize,
-    ) -> usize;
+        arg3: core::ffi::c_uint,
+    ) -> core::ffi::c_uint;
 }
 extern "C" {
     #[doc = "Get free heap size\n\nReturns:\n\n* free heap size in bytes\n\n"]
@@ -2096,7 +2097,7 @@ extern "C" {
     pub fn furi_thread_get_current_id() -> FuriThreadId;
 }
 extern "C" {
-    #[doc = "Get FuriThread instance for current thread\n\nReturns:\n\n* FuriThread*\n\n"]
+    #[doc = "Get FuriThread instance for current thread\n\nReturns:\n\n* pointer to FuriThread or NULL if this thread doesn't belongs to Furi\n\n"]
     pub fn furi_thread_get_current() -> *mut FuriThread;
 }
 extern "C" {
@@ -2136,8 +2137,8 @@ extern "C" {
     pub fn furi_thread_get_stdout_callback() -> FuriThreadStdoutWriteCallback;
 }
 extern "C" {
-    #[doc = "Set STDOUT callback for thread\n\nReturns:\n\n* true on success, otherwise fail\n\n# Arguments\n\n* `callback` - callback or NULL to clear\n\n"]
-    pub fn furi_thread_set_stdout_callback(callback: FuriThreadStdoutWriteCallback) -> bool;
+    #[doc = "Set STDOUT callback for thread\n\n# Arguments\n\n* `callback` - callback or NULL to clear\n\n"]
+    pub fn furi_thread_set_stdout_callback(callback: FuriThreadStdoutWriteCallback);
 }
 extern "C" {
     #[doc = "Write data to buffered STDOUT\n\nReturns:\n\n* size_t written data size\n\n# Arguments\n\n* `data` - input data\n* `size` - input data size\n\n"]
@@ -2353,6 +2354,15 @@ extern "C" {
 extern "C" {
     #[doc = "Is timer running\n\nReturns:\n\n* 0: not running, 1: running\n\n# Arguments\n\n* `instance` - The pointer to FuriTimer instance\n\n"]
     pub fn furi_timer_is_running(instance: *mut FuriTimer) -> u32;
+}
+pub type FuriTimerPendigCallback =
+    ::core::option::Option<unsafe extern "C" fn(context: *mut core::ffi::c_void, arg: u32)>;
+extern "C" {
+    pub fn furi_timer_pending_callback(
+        callback: FuriTimerPendigCallback,
+        context: *mut core::ffi::c_void,
+        arg: u32,
+    );
 }
 extern "C" {
     pub static _ctype_: [core::ffi::c_char; 0usize];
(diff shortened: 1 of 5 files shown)


crates/sys/src/bindings.rs
```diff
@@ -1936,8 +1937,8 @@ extern "C" {
     pub fn strlcpy(
         arg1: *mut core::ffi::c_char,
         arg2: *const core::ffi::c_char,
-        arg3: usize,
-    ) -> usize;
+        arg3: core::ffi::c_uint,
+    ) -> core::ffi::c_uint;
```
str4d, line 1941: Did you generate the bindings using a Windows machine? We've seen these flip-flop between `c_uint` and `usize` several times now as the bindings were updated.

crates/sys/src/bindings.rs
```diff
@@ -2096,7 +2097,7 @@ extern "C" {
     pub fn furi_thread_get_current_id() -> FuriThreadId;
 }
 extern "C" {
-    #[doc = "Get FuriThread instance for current thread\n\nReturns:\n\n* FuriThread*\n\n"]
+    #[doc = "Get FuriThread instance for current thread\n\nReturns:\n\n* pointer to FuriThread or NULL if this thread doesn't belongs to Furi\n\n"]
```
str4d, line 2100: I believe this shouldn't affect our `Thread` wrapper type's usage of `ptr::NonNull`, as my analysis here should still hold: https://github.com/flipperzero-rs/flipperzero/pull/67#discussion_r1193164706

crates/sys/src/bindings.rs
dcoles: Up to this point, yes, but I'm planning to use Linux (or WSL2) for any future updates.

crates/sys/src/bindings.rs
dcoles: While I can't think of any way to create a non-Furi thread, if the runtime doesn't perform a check, we should probably do so, if only to catch some future case where this call may return null for perhaps another reason.

Yahyaaddailam:
Update

crates/sys/src/bindings.rs
kitterion: As it turns out, this inconsistency is not because of linux vs windows but because of debian (and presumably all its derivatives like ubuntu) vs everybody else. Debian has like a million patches for llvm and one of them looks like this https://salsa.debian.org/pkg-llvm-team/llvm-toolchain/-/blob/14/debian/patches/23-strlcpy_strlcat_warning_removed.diff. Because of this patch debian's version of libclang.so (clang api library that is used by bindgen) does not consider `strlcpy` a builtin function. As a result, it becomes a regular user-defined function, and so its argument types get a `spelling = size_t` property in the clang AST that bindgen is able to parse and generate `usize` for them. 

I feel like saying that one must use a specific platform (debian linux in this case) to generate the bindings because of its quirks is wrong, and we should instead recommend a platform-independent way to do it. In this particuar case we could recommend using the upstream release artifacts, e.g. https://github.com/llvm/llvm-project/releases/tag/llvmorg-16.0.0. That's what I've been using and that's why it flip-flopped to `core::ffi::c_uint` this time. Thoughts on this?

There's another flip-flop incoming that starts with clang 16 (it's due to [this change](https://github.com/llvm/llvm-project/commit/9e956995db1fc7e792e3dfb3a465a52626195557)). It looks like this in the current pr
```diff
+pub type __gnuc_va_list = u32;
+pub type va_list = __gnuc_va_list;
-pub type va_list = u32;
```
Since I used the upstream release of clang 16, I hit both of them here. I can revert this one for now but eventually it will need to stay like this.

crates/sys/src/bindings.rs
kitterion: Switched from `NonNull::new_unchecked(...)` to `NonNull::new(...).expect(...)` for `furi_thread_get_current()`.

crates/sys/src/bindings.rs
dcoles: Great investigation tracking that down! It's a bit annoying, but at least it appears relatively harmless.

My ultimate goal is to eliminate the prebuilt `bindings.rs` (#41) due to the fact that it currently ties us to a specific Flipper Zero SDK version and makes it difficult to build against a dev branch. I'm about halfway there (https://github.com/flipperzero-rs/flipperzero/tree/build-sys), I just need to finish it off and then do a bunch of testing to make sure it works in multiple environments.

So I'm happy to leave it as it is, at least for the time being.

str4d: Given the above analysis, I've opened #72 to fix up the `Dockerfile` for generating bindings to be consistent with the non-Debian platforms.

str4d (approved):
ACK. I verified that I get the same bindings using `tools/Dockerfile`, modulo the two expected Debian-specific differences.

crates/sys/src/bindings.rs
kitterion: Much appreciated, but I was going to update the Dockerfile in this pr (or quickly after) and so I did 😅

docs/updating-sdk.md
```diff
@@ -33,10 +33,10 @@ Make sure to update the SDK details in [`README.md`] before making a new release
 
 Alternatively, you can generate `binding.rs` in an isolated env using Docker and the following command:
 
-From the root of the repository, to build the binding for the branch/tag `0.82.3` of the official SDK:
+From the root of the repository, to build the binding for the branch/tag `0.83.1` of the official SDK:
 
 ```shell
-docker run --rm -t $(docker build --build-arg BRANCH=0.82.3 -q -f tools/Dockerfile .) > crates/sys/src/bindings.rs
+docker run --rm $(docker build --build-arg BRANCH=0.83.1 -q -f tools/Dockerfile .) > crates/sys/src/bindings.rs
```
kitterion, line 39: I've removed `-t` from the docker command as it was causing `\r\n` line endings to be generated instead of just `\n`.

tools/Dockerfile
```diff
@@ -46,16 +45,15 @@ COPY . .
 
 WORKDIR /app/flipperzero-rs/tools
 
-RUN cargo run --bin generate-bindings /app/firmware/build/f7-firmware-D/sdk_headers/
+# We don't use debian's libclang. For details see https://github.com/flipperzero-rs/flipperzero/pull/70#discussion_r1199723419
+RUN env LIBCLANG_PATH=/app/clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04/lib cargo run --bin generate-bindings /app/firmware/build/f7-firmware-D/sdk_headers/
 
 RUN cp bindings.rs ../crates/sys/src
 
-RUN rustfmt --config newline_style=unix ../crates/sys/src/bindings.rs
```
kitterion, line 53: Explicitly running `rustfmt` is not necessary since bindgen runs `rustfmt` by default.
See https://rust-lang.github.io/rust-bindgen/code-formatting.html

kitterion:
I've updated the `tools/Dockerfile`, it should produce byte-for-byte identical `bindings.rs`.

tools/Dockerfile
str4d: I think this was added because for whatever reason, the first time the `Dockerfile` was run, everything was left unformatted on a single line. If that's no longer happening, great!

docs/updating-sdk.md
str4d: Aha, that's what the problem was! I think this is also what the `--config newline_style=unix` in the `rustfmt` line was trying to resolve.

str4d (approved):
ACK. I confirmed that I get identical bindings using the updated `Dockerfile`.

gmelodie:
Any updates on this? I tried making an app with @kitterion's branch but I still get the same API mismatch error.

str4d:
> So is this ready?

I believe so, although I now see that upstream bumped their SDK version again yesterday with firmware version 0.84.0-rc, so the upcoming firmware release will be incompatible again. From that perspective, I'm inclined to merge this (so people can work from the main branch), but wait the extra week or so until the next official firmware release, and then push out a crate version targeting that.

> Any updates on this? I tried making an app with @kitterion's branch but I still get the same API mismatch error.

@gmelodie that implies your Flipper Zero's firmware version is not either 0.82.3 (required for the version 0.9.0 crates) or 0.83.1 (required for this branch). Can you check what version you are running (either via qFlipper, or `Settings -> About -> last page)?

str4d:
0.84.1 was released 2 hours ago! I'll merge this and then get another PR up for that bump.

Issue #64, opened 2023-05-10
`rt`: Add framework for creating statics


str4d:
Normal Rust binaries will initialize statics on start, and then essentially leak the memory. This works fine because the memory is cleaned up during process exit. We can't assume the same as a FAP, so if we want to support "statics", we should provide a way in the runtime for initializing them before the user's app code runs, and freeing them after it exits.

dcoles:
In this case you're talking about statics that are stored on the heap (from [lazy_static](https://crates.io/crates/lazy_static), [once_cell](https://crates.io/crates/once_cell)), rather than static data stored in the `.rodata` right?

str4d:
Yep. People could still use one of those crates, but I believe it would get reported as a memory leak by the FZ tooling.

dcoles:
It might be possible to provide a factory function for [`OnceCell`](https://docs.rs/once_cell/1.17.1/once_cell/sync/struct.OnceCell.html) that registers the static for cleanup. Calling the [`take`](https://docs.rs/once_cell/1.17.1/once_cell/sync/struct.OnceCell.html#method.take) method takes ownership of the inner value allowing them to be dropped.

str4d:
We can't use `once_cell` because it requires `std` for `Mutex`. I'm instead planning on using `generic_once_cell`, which works for any mutex that implements `lock_api::RawMutex`. I've implemented the latter locally, and opened #74 with an `Instant` impl that I need.

dcoles:
> We can't use `once_cell` because it requires `std` for `Mutex`.

I think it's possible to use with `#[no-std]` as long as you provide a [`critical-section`](https://crates.io/crates/critical-section) implementation. I might take a look at that today, since a number of embedded crates are based around it.

PR #73, opened 2023-05-24
rt: Move thread waiter into a module

str4d:
This tidies up the `entry!` macro and makes it easier to add other state cleanup functions later.

### DIFF
--- a/crates/rt/src/lib.rs
+++ b/crates/rt/src/lib.rs
@@ -7,6 +7,7 @@
 
 pub mod manifest;
 pub mod panic_handler;
+mod thread;
 
 /// The C entry point.
 /// This just delegates to the user's Rust entry point.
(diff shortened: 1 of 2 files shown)


str4d:
#64 is an example of another change that would add a state cleanup function to the `entry!` macro.

str4d:
As a side-effect, this fixes a bug in the `entry!` macro where it requires `flipperzero-sys` to be a dependency of the macro user:
```
error[E0432]: unresolved import `flipperzero_sys`
  --> src/main.rs:11:1
   |
11 | entry!(main);
   | ^^^^^^^^^^^^ no external crate `flipperzero_sys`
   |
   = note: this error originates in the macro `entry` (in Nightly builds, run with -Z macro-backtrace for more info)

error: cannot determine resolution for the macro `sys::c_string`
  --> src/main.rs:11:1
   |
11 | entry!(main);
   | ^^^^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying macro imports
   = note: this error originates in the macro `entry` (in Nightly builds, run with -Z macro-backtrace for more info)
```

By moving the thread waiter into a module, `flipperzero-sys` is correctly resolved via the dependency of `flipperzero-rt`, and the macro user isn't affected (similarly to how we access it via `$crate::__macro_helpers::__sys` in the `flipperzero` crate to avoid this issue for `flipperzero` users).

dcoles (approved):
Looks good to me! 👍

PR #74, opened 2023-05-28
Implement `flipperzero::furi::time::{Instant, Duration}`

str4d:
We approximately follow the APIs of `std::time::{Instant, Duration}`, adjusting for the wrapping clock as necessary.

### DIFF
--- a/crates/flipperzero/src/furi/time.rs
+++ b/crates/flipperzero/src/furi/time.rs
@@ -1,5 +1,466 @@
-/// A placeholder `Instant` type. Currently unusable.
+use core::cmp::Ordering;
+use core::iter::Sum;
+use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Sub, SubAssign};
+
+use flipperzero_sys as sys;
+use ufmt::derive::uDebug;
+
+/// The maximum number of ticks that a [`Duration`] can contain for it to be usable with
+/// [`Instant`].
+const MAX_DURATION_TICKS: u32 = u32::MAX / 2;
+
+const NANOS_PER_SEC_F: f64 = 1_000_000_000_f64;
+const NANOS_PER_SEC: u64 = 1_000_000_000;
+const NANOS_PER_MILLI: u64 = 1_000_000;
+const NANOS_PER_MICRO: u64 = 1_000;
+const MILLIS_PER_SEC: u32 = 1_000;
+
+/// Converts the given number of nanoseconds to ticks.
+fn ns_to_ticks(nanos: u64) -> u64 {
+    let rate = unsafe { sys::furi_kernel_get_tick_frequency() };
+    if rate == MILLIS_PER_SEC {
+        // This can be up to around 2^45 ticks.
+        nanos / NANOS_PER_MILLI
+    } else {
+        // If `rate` is higher than `NANOS_PER_SEC_F` then this will overflow and be
+        // truncated. We assume that no Flipper Zero is clocked this fast.
+        ((f64::from(rate) / NANOS_PER_SEC_F) * nanos as f64) as u64
+    }
+}
+
+/// Converts the given number of ticks to nanoseconds.
+///
+/// The upper 2 bits of the return value will always be zero.
+fn ticks_to_ns(ticks: u32) -> u64 {
+    let rate = unsafe { sys::furi_kernel_get_tick_frequency() };
+    if rate == MILLIS_PER_SEC {
+        // This can be up to around 2^52 nanoseconds.
+        (ticks as u64) * NANOS_PER_MILLI
+    } else {
+        // This can be up to around 2^62 nanoseconds when `rate` is 1.
+        ((NANOS_PER_SEC_F / f64::from(rate)) * ticks as f64) as u64
+    }
+}
+
+/// A measurement of a wrapping clock. Opaque and useful only with [`Duration`].
+#[derive(Copy, Clone, Debug, uDebug, PartialEq, Eq, Hash)]
 pub struct Instant(pub(super) u32);
 
-/// A placeholder `Duration` type. Currently unusable.
+impl Instant {
+    /// Returns an instant corresponding to "now".
+    #[must_use]
+    pub fn now() -> Instant {
+        Instant(unsafe { sys::furi_get_tick() })
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `earlier` is later than `self`.
+    #[must_use]
+    pub fn duration_since(&self, earlier: Instant) -> Duration {
+        self.checked_duration_since(earlier)
+            .expect("earlier is later than self")
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one, or `None` if
+    /// that instant is later than this one.
+    #[must_use]
+    pub fn checked_duration_since(&self, earlier: Instant) -> Option<Duration> {
+        if self >= &earlier {
+            Some(Duration(self.0.wrapping_sub(earlier.0)))
+        } else {
+            None
+        }
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one, or zero
+    /// duration if that instant is later than this one.
+    #[must_use]
+    pub fn saturating_duration_since(&self, earlier: Instant) -> Duration {
+        self.checked_duration_since(earlier).unwrap_or_default()
+    }
+
+    /// Returns the amount of time elapsed since this instant.
+    ///
+    /// Due to the wrapping nature of the clock, there are several caveats on the value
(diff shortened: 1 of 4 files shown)


crates/flipperzero/src/furi/time.rs
```diff
@@ -0,0 +1,306 @@
...
+pub struct Instant(u64);
+
+impl Instant {
+    /// Returns an instant corresponding to "now".
+    #[must_use]
+    pub fn now() -> Instant {
+        let mut tick = 0;
+        let mut epoch = 0;
+
+        // Fetch `LAST_TICK` first because that is what updates most frequently.
```
dcoles, line 42: Full disclosure: I am far from being an expert on atomics.

This ordering looks backwards to me. I think it can result in a double update to `EPOC`.

For this example, let's consider `epoch` to be minutes and `tick` to be seconds (works easier in my head).

1. Initially `MINUTE = 0` and `SECOND = 59`
2. Thread 1 and Thread 2 load `SECOND` and both set `last_second = 59`
3. Thread 1 and Thread 2 call `now()`, one gets `second = 1` the other gets `second = 2`
4. Thread 1 increments `MINUTE` because `second < last_second`  (now `MINUTE = 1`)
5. Thread 2 increments `MINUTE` because `second < last_second`  (now `MINUTE = 2`)
6. Thread 1 stores `SECOND = 1`
7. Thread 2 tries to store `SECOND = 2`, but notices that `SECOND ≠ 59` so has to start the whole thing again
8. Thread 2 loads `SECOND` so its `last_second = 1`
9. Thread 2 calls `now()` and sets `second = 3`
10. Thread 2 does not increment `MINUTE` because `second >= last_second`
11. Thread 2 stores `SECOND = 3`
12. We finish with `MINUTE = 2` and `SECOND = 3`, but `MINUTE` should have only been `1`


I've currently thinking something like this:

```
let (epoch, tick) = loop {
    let last_epoch = EPOCH.load(Ordering::SeqCst);
    let last_tick = LAST_TICK.load(Ordering::SeqCst);
    let tick = sys::furi_get_tick();
    
    if LAST_TICK.compare_exchange(last_tick, tick, Ordering::SeqCst, Ordering::Relaxed).is_err();
        // Someone updated `LAST_TICK`, so we need to start again
        continue;
    }
    
    let epoch = if tick < last_tick {
        // Someone might beat us to updating `EPOCH`, but that's OK
        LAST_EPOCH.compare_exchange(last_epoch, last_epoch + 1, Ordering::SeqCst, Ordering::Relaxed).ok();
        last_epoch + 1
    } else {
        last_epoch
    };
    
    break (epoch, tick);
};
```

I'm pretty sure we can't have a weaker ordering than `SeqCst` since that would allow loading `LAST_TICK` before `EPOCH`. And while it might be possible to remove that retry-on-conflict loop, I really feel simpler is better when dealing with atomics.

str4d: Full disclosure: I am also far from being an expert on atomics 😅

My original plan had been to use a single `AtomicU64` to avoid all this trouble, but it's not available on this platform. The other technique I'd normally use is a mutex, but I'm using this to _build_ a mutex, so...

dcoles: Probably the best approach would be to use a single `u32` and `wrapping_sub`, which works through the wonders of modular arithmetic (e.g. `0x03 - 0xFE = 0x04`). This is the approach taken by [`embedded-time`](https://github.com/FluenTech/embedded-time) when a wrapping clock wraps around.

It's probably not worth the hassle of emulating an atomic-`u64` (we might even be hurting performance with atomic operations). Otherwise, I'd want to find some existing atomic-`u64` implementation and base it upon that.

Alternatively, it might not be a bad idea to consider something like [`fugit`](https://crates.io/crates/fugit) or [`embedded-time`](https://crates.io/crates/embedded-time) rather than implementing our own `Instant`.

str4d: > Probably the best approach would be to use a single `u32` and `wrapping_sub`, which works through the wonders of modular arithmetic (e.g. `0x03 - 0xFE = 0x04`). This is the approach taken by [`embedded-time`](https://github.com/FluenTech/embedded-time) when a wrapping clock wraps around ([FluenTech/embedded-time#5](https://github.com/FluenTech/embedded-time/issues/5)).

That's a good point. I was trying to replicate `std::time::Instant` which is monotonically non-decreasing and thus never wraps, but if we're okay with only dealing in durations of max length `u32::MAX / 2` ticks (around 23 days at the default tick frequency of 1000 Hz) then a wrapping `Instant` would be fine.

> Alternatively, it might not be a bad idea to consider something like [`fugit`](https://crates.io/crates/fugit) or [`embedded-time`](https://crates.io/crates/embedded-time) rather than implementing our own `Instant`.

Both `fugit` and `embedded-time` bake the scaling fraction in as a constant, which isn't compatible with the runtime-only `sys::furi_kernel_get_tick_frequency`. However, we _could_ make the assumption that we are only working with official firmware, and bake in the tick frequency ourselves (with a test to ensure we notice if upstream changes it). This would block compatibility with any custom firmware that changes the kernel tick frequency.

As far as APIs go, both of them eschew `core::time::Duration` in favour of their own type, for varying reasons:
- `fugit` wants to do everything at compile time. It looks pretty optimised for these range conversions, but has a rather verbose and large API as a result of the compile-time mechanics.
- `embedded-time` has a somewhat more organised API that would be easier to expose, and mainly avoids `core::time::Duration` because of the inefficiency in how it represents intervals (whereas `embedded-time` does the same thing as `fugit` using a scaling ratio, except at runtime). But there's a weird asymmetry where durations obtained from `Instant` are via a `duration::Generic` type, but applying duration offsets to `Instant` is only supported via fixed-resolution `duration::*` types, forcing you to convert from `duration::Generic` to e.g. `duration::Milliseconds` if you need to round-trip.

Of the two I prefer `embedded-time`, but its developer also hasn't maintained it in a year and a half (vs `fugit` having infrequent but minimal maintenance). I did submit a review on one of their PRs for fixing a breaking bug in rapid `Instant` comparisons; maybe that will bump it into the maintainer's view.

dcoles: > That's a good point. I was trying to replicate `std::time::Instant` which is monotonically non-decreasing and thus never wraps, but if we're okay with only dealing in durations of max length `u32::MAX / 2` ticks (around 23 days at the default tick frequency of 1000 Hz) then a wrapping `Instant` would be fine.

I think that should be fine, though maybe we should note this in the documentation. I think it will even handle up to a full `u32::MAX - 1` ticks (e.g. `0xFFFFFF7Eu32 - 0xFFFFFF7Fu32 = 0xFFFFFFFEu32`).

> Both `fugit` and `embedded-time` bake the scaling fraction in as a constant, which isn't compatible with the runtime-only `sys::furi_kernel_get_tick_frequency`.

That's a good point. I guess the intent is that `.fap` binaries should be portable across future devices. If so, that probably leaves us with our own `Instant` type that does a runtime conversion from ticks to `Duration`.

str4d: > I think it will even handle up to a full `u32::MAX - 1` ticks (e.g. `0xFFFFFF7Eu32 - 0xFFFFFF7Fu32 = 0xFFFFFFFEu32`).

This would work, but it prevents an `Ord` impl. The `embedded-time` crate enforces a `u32::MAX / 2` limit on `Duration` so that it can determine ordering based on which side of that boundary `a.wrapping_sub(b)` falls on.

dcoles: > This would work, but it prevents an `Ord` impl. The embedded-time crate enforces a `u32::MAX / 2` limit on Duration so that it can determine ordering based on which side of that boundary `a.wrapping_sub(b)` falls on.
Oh, interesting. That's pretty clever. I saw some code using `i32::wrapping_sub` which looked quite odd, but just realized it has the nice property that you get a signed delta (i.e. is it earlier or later). But using a `u32` is far less confusing when talking about a tick counter.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/bd2871f29eb19867bd8f40d6b1aa7884774e1b85..8eb81c87989a29d485c930f9721b9c1d43374f47).

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/8eb81c87989a29d485c930f9721b9c1d43374f47..fddba1c00d9efebd9588083f462c4f9d97fb7e61) to rework the implementation of `Instant` to be wrapping, and added an implementation of `Duration`.

crates/flipperzero/src/furi/time.rs
str4d: The implementation should now be wrapping.

dcoles (approved):
Looks good to me! 👍

PR #75, opened 2023-05-31
Rewrite `Mutex` type using the `lock_api` crate

str4d:


crates/flipperzero/src/furi/sync.rs
```diff
@@ -1,82 +1,108 @@
...
         }
     }
 }
 
-// `UnsendUnsync` is actually a bit too strong.
-// As long as `T` implements `Sync`, it's fine to access it from another thread.
-unsafe impl<T: ?Sized + Sync> Sync for MutexGuard<'_, T> {}
+unsafe impl RawMutexTimed for FuriMutex {
+    type Duration = u32; // TODO replace with `Duration` type that hides ticks.
+    type Instant = u32; // TODO replace with `Instant` type that hides ticks.
```
str4d, line 88: I opened this PR now so it can be reviewed, but we probably don't want to merge it before we have decided on a solution for these types.

crates/flipperzero/src/furi/sync.rs
```diff
@@ -1,82 +1,108 @@
...
+    const INIT: Self = FuriMutex::new();
+    type GuardMarker = GuardNoSend;
 
-    fn deref(&self) -> &Self::Target {
-        unsafe { &*self.0.data.get() }
+    fn lock(&self) {
+        // `INCLUDE_vTaskSuspend` is set to 1 in the Flipper Zero's FreeRTOS config, so as
+        // long as this timeout value matches `portMAX_DELAY` in the FreeRTOS config, this
+        // will block indefinitely as intended.
+        assert!(self.try_acquire(u32::MAX));
```
dcoles, line 70: There is a `FuriWaitForever`, but sadly it doesn't seem to be exposed via the SDK. ☹️

crates/flipperzero/src/furi/sync.rs
```diff
@@ -1,82 +1,108 @@
...
     }
-}
 
-impl<T> DerefMut for MutexGuard<'_, T> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        unsafe { &mut *self.0.data.get() }
+    fn try_lock(&self) -> bool {
+        // Poll first so we don't block at all if we cannot acquire.
+        self.try_acquire(0) && self.try_acquire(1)
```
dcoles, line 75: Calling `xSemaphoreTake` with a timeout of 0 will acquire the lock if possible, so there's no need to call it a second time. The wording of the documentation is a bit ambiguous, but the behaviour is clarified in [this support thread](https://www.freertos.org/FreeRTOS_Support_Forum_Archive/September_2015/freertos_xSemaphoreTake_macro_with_delay_0_cannot_work_ec38d65fj.html).

crates/flipperzero/src/furi/sync.rs
dcoles: In the short term, you could create a `Duration(u32)` and `Instant(u32)` new-type so that the implementation is opaque.

This can then be replaced with real `Duration` and `Instant` types once we are happy with an approach.

crates/flipperzero/src/furi/sync.rs
```diff
@@ -1,82 +1,108 @@
...
-unsafe impl<T: ?Sized + Sync> Sync for MutexGuard<'_, T> {}
+unsafe impl RawMutexTimed for FuriMutex {
+    type Duration = u32; // TODO replace with `Duration` type that hides ticks.
+    type Instant = u32; // TODO replace with `Instant` type that hides ticks.
+
+    fn try_lock_for(&self, timeout: Self::Duration) -> bool {
+        self.try_acquire(match timeout {
+            // Add 1 tick to avoid the polling behaviour.
+            0 => 1,
+            n => n,
```
dcoles, line 94: Based on the comment about `xSemaphoreTake` with a timeout of 0, I don't believe there's need for any special behaviour here.

crates/flipperzero/src/furi/sync.rs
str4d: Heh, that is _not_ clear from the documentation. This is also how I originally assumed it would work, before reading said documentation and changing to the above approach 🤣 I'll change it back.

crates/flipperzero/src/furi/time.rs
```diff
@@ -0,0 +1,5 @@
+/// A placeholder `Instant` type. Currently unusable.
+pub struct Instant(pub(super) u32);
+
+/// A placeholder `Duration` type. Currently unusable.
+pub struct Duration(pub(super) u32);
```
str4d, line 5: These being opaque with no operators effectively (temporarily) disables the "explicit timeout" `Mutex` APIs for now, while still keeping them implemented. Then once these are replaced by #74 the APIs will become usable.

dcoles (approved):
Looks good to me!

PR #79, opened 2023-06-07
test: Support `#[cfg(..)]` attributes on test methods

str4d:
This is primarily to enable tests behind `#[cfg(feature = "alloc")]`, but may also be useful if we refactor to have services behind feature flags (https://github.com/flipperzero-rs/flipperzero/pull/29#discussion_r1059691569).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
+### Changed
+- `flipperzero_test::tests` now allows `#[cfg(..)]` attributes on test methods.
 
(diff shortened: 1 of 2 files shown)


JarvisCraft:
By the way, can't this approach be generalizzed to all attributes instead of this specific one?

str4d:
> By the way, can't this approach be generalizzed to all attributes instead of this specific one?

It will need to be on a case-by-case basis. We already preserve all attributes on the tests themselves, so things like `allow` should work fine in most cases. The reason `cfg` needs special handling is because the test names are stored in a slice so the runner can find them, and `cfg` makes their compilation conditional, so we need to apply the same conditions to the slice entry.

JarvisCraft (approved):
Thanks for the explanation!

PR #80, opened 2023-06-08
Dialogs app file browser wrapper

mogenson:
Add a safe Rust wrapper for the dialog file browser methods and the
DialogsFileBrowserOptions type. Rename the DialogsApp::show() method
to show_message() and create new show_file_browser() method. Add a
getter method to FuriString to provide access to the inner
sys::FuriString via a raw pointer because a sys::FuriString is needed by
a few of the Furi C API functions. Extend the storage example to open a
file browser and have the user select the file that was just written to
the file system.

Note: The Furi C struct is named DialogsFileBrowserOptions, but all
other dialog types and functions are singular instead of plural. The
Rust type is named DialogFileBrowserOptions.

The DialogFileBrowserOptions type uses setter methods that return Self
so they can be chained together in the builder pattern. The
DialogFileBrowserOptions includes a Rust unit type phantom data member.
The type or size does not really matter, but PhantomData does require a
sized type. We just need one struct member to carry the lifetime
parameter so that CStr and Icon references are not dropped until the
struct goes out of scope.

### DIFF
--- a/crates/flipperzero/examples/storage.rs
+++ b/crates/flipperzero/examples/storage.rs
@@ -13,6 +13,8 @@ extern crate flipperzero_alloc;
 
 use core::ffi::CStr;
 
+use flipperzero::dialogs::{DialogFileBrowserOptions, DialogsApp};
+use flipperzero::furi::string::FuriString;
 use flipperzero::io::*;
 use flipperzero::println;
 use flipperzero::storage::*;
@@ -23,10 +25,11 @@ entry!(main);
 
 fn main(_args: *mut u8) -> i32 {
     // First, we'll create a file on the SD card and write "Hello, Rust!" to it.
+    let path = CStr::from_bytes_with_nul(b"/ext/hello-rust.txt\0").unwrap();
     let file = OpenOptions::new()
         .write(true)
         .create_always(true)
-        .open(CStr::from_bytes_with_nul(b"/ext/hello-rust.txt\0").unwrap());
+        .open(path);
 
     match file {
         Ok(mut handle) => {
@@ -37,21 +40,30 @@ fn main(_args: *mut u8) -> i32 {
         Err(e) => println!("couldn't open path: {}", e),
     }
 
-    // Now, we'll open it and read it back.
-    let mut buffer: [u8; 16] = [0; 16];
-    let file = OpenOptions::new()
-        .read(true)
-        .open(CStr::from_bytes_with_nul(b"/ext/hello-rust.txt\0").unwrap());
+    // Next, we'll open a file browser dialog and let the user select the file.
+    let mut dialogs_app = DialogsApp::open();
+    let file_browser_options = DialogFileBrowserOptions::new().set_hide_ext(false);
+    let mut start_path = FuriString::from(path);
+    let result_path =
+        dialogs_app.show_file_browser(Some(&mut start_path), Some(&file_browser_options));
+    if let Some(result_path) = result_path {
+        println!("file selected {}", result_path);
+        let path = result_path.as_c_str();
 
-    match file {
-        Ok(mut handle) => match handle.read(&mut buffer) {
-            Ok(n) => println!("Read from file: {:?}", &buffer[..n]),
-            Err(e) => println!("couldn't read from file: {}", e),
-        },
-        Err(e) => println!("couldn't open path: {}", e),
+        // Now, we'll open it and read it back.
+        let mut buffer: [u8; 16] = [0; 16];
+        let file = OpenOptions::new().read(true).open(path);
+
+        match file {
+            Ok(mut handle) => match handle.read(&mut buffer) {
+                Ok(n) => println!("Read from file: {:?}", &buffer[..n]),
+                Err(e) => println!("couldn't read from file: {}", e),
+            },
+            Err(e) => println!("couldn't open path: {}", e),
+        }
+    } else {
+        println!("no file selected");
     }
 
-    0
-
-    // File is synchronized and closed when `f` goes out of scope.
+    0 // File is synchronized and closed when `file` goes out of scope.
 }
--- a/crates/flipperzero/src/dialogs/mod.rs
+++ b/crates/flipperzero/src/dialogs/mod.rs
@@ -3,13 +3,14 @@
 #[cfg(feature = "alloc")]
 use alloc::ffi::CString;
 
-use core::ffi::{c_char, CStr};
+use core::ffi::{c_char, c_void, CStr};
 use core::marker::PhantomData;
 use core::ptr::{self, NonNull};
 
 use flipperzero_sys as sys;
-use flipperzero_sys::furi::UnsafeRecord;
+use sys::{c_string, furi::UnsafeRecord};
 
+use crate::furi::string::FuriString;
 use crate::gui::canvas::Align;
 
 const RECORD_DIALOGS: *const c_char = sys::c_string!("dialogs");
@@ -26,6 +27,12 @@ pub struct DialogsApp {
 pub struct DialogMessage<'a> {
     data: NonNull<sys::DialogMessage>,
     _phantom: PhantomData<&'a CStr>,
+}
+
+/// A dialog file browser options.
+pub struct DialogFileBrowserOptions<'a> {
+    data: sys::DialogsFileBrowserOptions,
+    _phantom: PhantomData<&'a ()>,
 }
 
 /// Button pressed on a dialog.
@@ -45,11 +52,37 @@ impl DialogsApp {
     }
 
     /// Displays a message.
-    pub fn show(&mut self, message: &DialogMessage) -> DialogMessageButton {
+    pub fn show_message(&mut self, message: &DialogMessage) -> DialogMessageButton {
         let button_sys =
             unsafe { sys::dialog_message_show(self.data.as_ptr(), message.data.as_ptr()) };
 
         DialogMessageButton::from_sys(button_sys).expect("Invalid button")
+    }
+
+    /// Displays a file browser.
+    ///  - path is a optional preselected file path
+    ///  - options are optional file browser options
+    pub fn show_file_browser(
+        &mut self,
+        path: Option<&mut FuriString>,
+        options: Option<&DialogFileBrowserOptions>,
+    ) -> Option<FuriString> {
+        let mut result_path = FuriString::new();
+        // path will be unmodified but needs to be a valid FuriString.
+        // We can reuse the empty result_path if path is not provided.
+        let path = path.unwrap_or(&mut result_path).as_mut_ptr();
+        let options = options
+            .map(|opts| &opts.data as *const sys::DialogsFileBrowserOptions)
+            .unwrap_or(ptr::null());
+        unsafe {
+            sys::dialog_file_browser_show(
+                self.data.as_ptr(),
+                result_path.as_mut_ptr(),
+                path,
+                options,
+            )
+        }
+        .then_some(result_path)
     }
 }
 
@@ -164,6 +197,73 @@ impl DialogMessageButton {
     }
 }
 
+impl<'a> DialogFileBrowserOptions<'a> {
+    /// Creates a new dialog file browser options and initializes to default values.
+    pub fn new() -> Self {
+        Self {
+            // default values from sys::dialog_file_browser_set_basic_options()
+            data: sys::DialogsFileBrowserOptions {
+                extension: c_string!("*"),
+                base_path: ptr::null(),
+                skip_assets: true,
+                hide_dot_files: false,
+                icon: ptr::null(),
+                hide_ext: true,
+                item_loader_callback: None,
+                item_loader_context: ptr::null_mut(),
+            },
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Set file extension to be offered for selection.
+    pub fn set_extension(mut self, extension: &'a CStr) -> Self {
+        self.data.extension = extension.as_ptr();
+        self
+    }
+
+    /// Set root folder path for navigation with back key.
+    pub fn set_base_path(mut self, base_path: &'a CStr) -> Self {
+        self.data.base_path = base_path.as_ptr();
+        self
+    }
+
+    /// Set file icon.
+    pub fn set_icon(mut self, icon: &'a sys::Icon) -> Self {
+        self.data.icon = icon as *const sys::Icon;
+        self
+    }
+
+    /// Do not show assets folder if true.
+    pub fn set_skip_assets(mut self, skip_assets: bool) -> Self {
+        self.data.skip_assets = skip_assets;
+        self
+    }
+
+    /// Hide dot files if true.
+    pub fn set_hide_dot_files(mut self, hide_dot_files: bool) -> Self {
+        self.data.hide_dot_files = hide_dot_files;
+        self
+    }
+
+    /// Hide extensions for files if true.
+    pub fn set_hide_ext(mut self, hide_ext: bool) -> Self {
+        self.data.hide_ext = hide_ext;
+        self
+    }
+
+    /// Set callback function and context for providing custom icon and entry name.
+    pub fn set_item_loader_callback(
+        mut self,
+        callback: sys::FileBrowserLoadItemCallback,
+        context: *mut c_void,
+    ) -> Self {
+        self.data.item_loader_callback = callback;
+        self.data.item_loader_context = context;
+        self
+    }
+}
+
 /// Displays a simple dialog.
 #[cfg(feature = "alloc")]
 pub fn alert(text: &str) {
@@ -175,5 +275,5 @@ pub fn alert(text: &str) {
     message.set_text(&text, 0, 0, Align::Left, Align::Top);
     message.set_buttons(None, Some(BUTTON_OK), None);
 
-    dialogs.show(&message);
+    dialogs.show_message(&message);
 }
--- a/crates/flipperzero/src/furi/string.rs
+++ b/crates/flipperzero/src/furi/string.rs
@@ -84,6 +84,13 @@ impl FuriString {
         unsafe { CStr::from_ptr(self.as_c_ptr()) }
     }
 
+    /// Raw pointer to the inner sys::FuriString
+    #[inline]
+    #[must_use]
+    pub(crate) fn as_mut_ptr(&mut self) -> *mut sys::FuriString {
+        self.0.as_ptr()
+    }
+
     /// Appends a given `FuriString` onto the end of this `FuriString`.
     #[inline]
     pub fn push_string(&mut self, string: &FuriString) {
--- a/crates/flipperzero/examples/dialog.rs
+++ b/crates/flipperzero/examples/dialog.rs
@@ -45,7 +45,7 @@ fn main(_args: *mut u8) -> i32 {
         Some(CStr::from_bytes_with_nul(b"Scissor\0").unwrap()),
     );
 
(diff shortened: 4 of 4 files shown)


JarvisCraft (changes requested):

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -45,12 +53,42 @@ impl DialogsApp {
...
+                self.data.as_ptr(),
+                result_path.as_ptr(),
+                path_ptr,
+                options_ptr,
+            )
+        } {
+            Some(result_path)
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
mogenson: Thanks for the tip! I will change.

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -164,6 +202,84 @@ impl DialogMessageButton {
     }
 }
 
+impl<'a> DialogFileBrowserOptions<'a> {
+    /// Creates a new dialog file browser options and initializes to default values.
+    pub fn new(extension: Option<&'a CStr>, icon: Option<&'a sys::Icon>) -> Self {
+        let mut dialog_file_browser_options = Self {
+            data: unsafe { mem::zeroed() },
```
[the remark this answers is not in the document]
mogenson: I was following the pattern found in C code of creating an uninitialized struct: `DialogsFileBrowserOptions = {};` and passing it to `dialog_file_browser_set_basic_options()` to initialize.

I can initialize all the struct values how `dialog_file_browser_set_basic_options()` does in Rust and avoid the call into C. The options can then be customized with the setter methods. I think I'll remove the `Default` trait implementation because `new()` will be the default.

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -164,6 +202,84 @@ impl DialogMessageButton {
...
+        let ext_ptr = extension
+            .unwrap_or(Default::default())
+            .as_ptr();
+        let icon_ptr = match icon {
+            Some(icon) => icon as *const sys::Icon,
+            None => ptr::null(),
+        };
+        unsafe {
+            sys::dialog_file_browser_set_basic_options(
+                &mut dialog_file_browser_options.data as *mut sys::DialogsFileBrowserOptions,
```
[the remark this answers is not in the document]
mogenson: This shouldn't be moving the address of `dialog_file_browser_options` we are passing a raw pointer to the inner `sys::DialogsFileBrowserOptions` for the C function to populate. This is just a helper function that sets things like `options->icon = icon;` in C instead of us assigning the variable directly in Rust. Anyways, I'm removing this FFI call and now doing all the default value assignments in Rust.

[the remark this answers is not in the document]
JarvisCraft: Actually, we can use `dialog_file_browser_set_basic_options` safely (from API standpoint) which would allow us to not copy the defaults (which may be important if the deafult change in C impl) and yet keep soundness: you could use `MaybeUninit` for stack-allocating the structure and then passing it to `dialog_file_browser_set_basic_options`.

crates/flipperzero/src/dialogs/mod.rs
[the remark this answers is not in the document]
mogenson: Check out the current change. By the way, in the Furi C code, some pointer values are allowed to be NULL. For example, the `base_path`, `icon, and `item_loader_context` can all be NULL, these are checked in the C code. The `extension` member has to be a valid string. When calling `sys::dialog_file_browser_show()`, you can also pass NULL for the file browser options.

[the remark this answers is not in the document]
JarvisCraft: My apologize for a bit chaotic review :heart:

crates/flipperzero/src/dialogs/mod.rs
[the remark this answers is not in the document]
mogenson: Ah, well now that you've seen things both ways do you have a strong preference? I also just noticed that `dialog_file_browser_set_basic_options()` doesn't touch the `hide_dot_files` member. It seems to rely on a struct being zeroed and zero being a falsey value in C.

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -45,12 +52,34 @@ impl DialogsApp {
...
+        let result_path = FuriString::new();
+        // path will be unmodified but needs to be a valid FuriString.
+        // We can reuse the empty result_path if path is not provided.
+        let path = path.unwrap_or(&result_path).as_ptr();
+        let options = match options {
+            Some(opts) => &opts.data as *const sys::DialogsFileBrowserOptions,
+            None => ptr::null(),
+        };
+        unsafe {
+            sys::dialog_file_browser_show(self.data.as_ptr(), result_path.as_ptr(), path, options)
```
JarvisCraft, line 79: I fell like this is unsound:

1. `options` is passed by `&` and the return type has no lifetimes;
2. you convert `&opts.data` to a `const`-pointer;
3. you return from the function thus the original `&DialogFileBrowserOptions` can now be moved.

Now if the API user moves the `DialogFileBrowserOptions` to which he pointed in the constructor, it will include moving its `data` field which will invalidate the `options` pointer passed to `dialog_file_browser_show`

JarvisCraft (changes requested):

crates/flipperzero/src/dialogs/mod.rs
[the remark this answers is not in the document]
JarvisCraft: That's a really interesting question.

It may just be secretly unsound in flipperzero-firmware since their `DialogsFileBrowserOptions browser_options` initialization does not guarantee zero-filling (AFAIK).

Let's keep rust-based initialization as for now, since it is easier to speculate on its safety as for now. But I will ask firmware maintainers if it's just an overlooked field in their initialization and we can ask them to make `dialog_file_browser_set_basic_options()` always fully initialize the instance so that we can rely on it (and in this case we will add a future PR to replace current approach with `MaybeUnit`-based).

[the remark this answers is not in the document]
JarvisCraft: I've [asked](https://t.me/flipperzero/103925/177467) them but I suggest that we don't wait fr the response (since it also requires filing a PR to firmware, a new release of firware and update of our bindings) and use Rust-approach by now.

crates/flipperzero/src/dialogs/mod.rs
mogenson: I think this is ok, but I'm not an expert on Rust lifetimes.
1. `options` is only used as a reference and is unmodified by this Rust function or the `dialog_file_browser_show` C function. The C function is blocking, when it returns, we are done with the `options` and it can be dropped by the caller when it goes out of scope. So I don't think we need to move `options` into `show_file_browser` or hold onto a reference.
2. Yep, `dialog_file_browser_show` wants a const pointer to a `DialogsFileBrowserOptions` (the inner C struct). The const designates that C won't modify this struct and we can inspect the C code to know that it won't hold onto this pointer after the `dialog_file_brower_show` function returns.
3. The caller can do whatever they want to the original `DialogFileBrowserOptions` after this function, no data will be modified from underneath them.

JarvisCraft: Hmm, I haven't thought that `dialog_file_browser_show` blocks until it has done all its work.

In this case, yes, everyhting seems fine :rocket:

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -28,6 +29,12 @@ pub struct DialogMessage<'a> {
     _phantom: PhantomData<&'a CStr>,
 }
 
+/// A dialog file browser options.
+pub struct DialogFileBrowserOptions<'a> {
+    data: sys::DialogsFileBrowserOptions,
+    _phantom: PhantomData<&'a c_void>,
```
JarvisCraft, line 35: Extra small detail: I guess `()` is a more common type to be used in `PhantomReference`:

```suggestion
    _phantom: PhantomData<&'a ()>,
```

JarvisCraft (changes requested):

mogenson:
@JarvisCraft are you ok with merging this now?

JarvisCraft:
I actually am not the one with push permissions, so I summon @dcoles and @str4d for this purpose :)

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -45,12 +52,34 @@ impl DialogsApp {
...
+    pub fn show_file_browser(
+        &mut self,
+        path: Option<&FuriString>,
+        options: Option<&DialogFileBrowserOptions>,
+    ) -> Option<FuriString> {
+        let result_path = FuriString::new();
+        // path will be unmodified but needs to be a valid FuriString.
+        // We can reuse the empty result_path if path is not provided.
+        let path = path.unwrap_or(&result_path).as_ptr();
+        let options = match options {
```
dcoles, line 74: You may wish to consider:
```rust
let options = options.map(|opts| &opts.data as *const sys::DialogsFileBrowserOptions).unwrap_or(ptr::null());
```

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -164,6 +193,73 @@ impl DialogMessageButton {
     }
 }
 
+impl<'a> DialogFileBrowserOptions<'a> {
+    /// Creates a new dialog file browser options and initializes to default values.
+    pub fn new() -> Self {
+        Self {
+            // default values from sys::dialog_file_browser_set_basic_options()
+            data: sys::DialogsFileBrowserOptions {
+                extension: CStr::from_bytes_until_nul(b"*\0").unwrap().as_ptr(),
```
dcoles, line 202: You can also use `sys::c_string!` here too.

crates/flipperzero/src/furi/string.rs
```diff
@@ -84,6 +84,13 @@ impl FuriString {
         unsafe { CStr::from_ptr(self.as_c_ptr()) }
     }
 
+    /// Raw pointer to the inner sys::FuriString
+    #[inline]
+    #[must_use]
+    pub fn as_ptr(&self) -> *mut sys::FuriString {
```
dcoles, line 90: This should take `&mut self` and be called `as_mut` since this effectively allows you mutable access to the string data.

crates/flipperzero/examples/storage.rs
```diff
@@ -23,10 +25,11 @@ entry!(main);
 
 fn main(_args: *mut u8) -> i32 {
     // First, we'll create a file on the SD card and write "Hello, Rust!" to it.
+    let path = CStr::from_bytes_with_nul(b"/ext/hello-rust.txt\0").unwrap();
```
dcoles, line 28: We should use the `sys::c_string!` macro here (it automatically adds a b"\0").

dcoles (changes requested):
Hi @mogenson,

Thanks for the PR! This looks great.

Most of my feedback is just minor style things, though I'd say the `as_ptr` returning a `*mut sys::FuriString` is important to address before I can merge this.

crates/flipperzero/src/furi/string.rs
JarvisCraft: I would disagree. `core` and `std` types do not require `mut` on `self` since the raw pointer itself cannot be used in any bad way, while its dereferenceing is anyway `unsafe`.

Also our existing APIs use `&self`, for example `UnsafeRecord::as_ptr(&self)`

crates/flipperzero/src/dialogs/mod.rs
mogenson: Thanks for the tip. Will do!

crates/flipperzero/examples/storage.rs
mogenson: Good to know. Thanks.

crates/flipperzero/src/furi/string.rs
mogenson: I was a bit unsure about this method because it's just a wrapper for the `as_ptr()` method of the core `NonNull` type. `NonNull<T>::as_ptr()` takes `self` and returns an `*mut T`.

str4d: This method absolutely needs to take `&mut self`. AFAICT the only place this PR uses this method is `sys::dialog_file_browser_show`, which mutates the provided `sys::FuriString` to store the selected path. From the Rust perspective, this is mutating the `FuriString` and therefore requires exclusive access.

I also intentionally did not add this method to the public API, because I do not want it to exist there; it enables violating the safety requirements relied on by the `FuriString` struct. We should instead be crafting safe APIs that wrap the FFI methods taking `*mut sys::FuriString`, inside which we can call this in a crate-private setting where we know the safety requirements are satisfied.

Finally, we should name this `as_mut_ptr` to match equivalent APIs in the Rust stdlib ([`slice::as_mut_ptr`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.as_mut_ptr), [`Vec::as_mut_ptr`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.as_mut_ptr), [`str::as_mut_ptr`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.as_mut_ptr)). I'd personally also prefer that this method be `unsafe`, though technically obtaining this kind of reference is safe Rust (it's just using said reference that is unsafe), so I'm not going to require that.
```suggestion
    pub(crate) fn as_mut_ptr(&mut self) -> *mut sys::FuriString {
```

crates/flipperzero/examples/storage.rs
dcoles: Oh, sorry. I don't know how I missed that. This is fine.

crates/flipperzero/src/furi/string.rs
dcoles: @JarvisCraft As far as I'm aware methods like [`str::as_mut_ptr`](https://doc.rust-lang.org/std/primitive.str.html#method.as_mut_ptr) all need `&mut self` (or `self` so they consume the original object).

dcoles: @str4d I think there's always going to be a need for folks to be able to get access to the underlying `*sys::FuriString` for any APIs that we haven't wrapped (e.g. maybe a third party Flipper Zero C library).

> I do not want it to exist there; it enables violating the safety requirements relied on by the `FuriString` struct

I'm trying to think of which safety requirements you could violate. My understanding was that `FuriString` was just a thin wrapper around `NonNull<sys::FuriString>`, so as long as you don't violate any standard ownership rules (e.g. calling `furi_string_free` on a string that you aren't the owner of), then it can be used safely.

I did see that our `FuriString` has the property "no nul bytes in the middle", but that might not be strictly necessary (if `m-string` allows it). While `CString` forbids an interior nul byte (I guess so the `into_raw`/`from_raw` pair works), `CStr` does not have that requirement.

Let me know if I'm missing something (which is well possible here).

JarvisCraft: In my opinion, adding `&mut` is not the right choice here.

It does not add to the safety since you [can](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ae5ff565df8bdd967ca844eeaefb1d78) safely convert `*const T` to `*mut T`. At the same time, this requirement would add a faulty feeling in safety to its user making him think that something is guaranteed by it (which is false).

JarvisCraft: From the other side, I like the crate-local access modifier so I generally like @str4d's suggestion.

JarvisCraft: Actually, you can propose this change to flipperdevices/flipperzero-firmware. I've done similar job with adding `const`-qualifiers and they were fine with merging them.

dcoles: > Actually, `NonNull` is [`impl Copy`](https://doc.rust-lang.org/src/core/ptr/non_null.rs.html#701) so it does not effectively consume it.

Oh! I did not realize that was the case. OK. That makes more sense. Also I see `NonNull<T>` is defined as being the same as `*mut T`, so there's no need for a `mut self`, because `self` *is* a pointer-to-mut-T.

> In my opinion, adding `&mut` is not the right choice here.

I see this is analogous to `String::as_ptr(&self) -> *const u8` and `String::as_mut_ptr(&mut self) -> *mut u8`. Sure, you can freely cast back and forth between `*const` and `*mut` (since Rust won't let you actually _do_ anything with these pointers in safe code). However if you (the user) violate mutability rules (e.g. by creating a `mut*` to an immutable object and then mutatating it in `unsafe` code), then that is not Rust's fault. You, the user, are responsible for upholding Rust's mutability rules in `unsafe` code.

`const*` doesn't really mean all that much in Rust (or C). It's essentially type-checked documentation. About the only thing it does in Rust is [prevents you from performing mutable operations when you dereference it](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b95ce199ad8d90a13522a3c2a194dcb4).

Is it useless? No, it's still helpful to have the compiler act as a safety. The `furi_string_` methods are all const-correct and will stop you if you try to mutate something through a `const*`. If you take off the safety and point it at your foot, well you're in unsafe code, so neither Rust or C is going to stop you.

This would all be far less confusing if we could have done in-place/`#[repr(transparent)]` allocation of `FuriString` (flipperdevices/flipperzero-firmware#2461). Then it's super clear that you can't get a `*FuriString` that you can legally write through without having a `mut self` first.

> From the other side, I like the crate-local access modifier so I generally like @str4d's suggestion.

OK, I have no problem with keeping this `pub(crate)`.

dcoles: > Maybe we can include a note in the comment doc that although this method needs a mutable reference, the starting file path will not be modified. It would be better if the `sys::dialog_show_file_browser()` C function took a `*const FuriString` for path instead of `*FuriString`.

This is one totally legitimate case for `*const T as *mut T`. It's fine to call `FuriString::as_ptr() -> *const sys::FuriString` and then cast it to a `*mut sys::FuriString` so long as you don't modify memory through this pointer.

Ultimately what matters (for correct compilation) is "does `sys::dialog_show_file_browser()` actually mutate the data behind the `path` pointer?", not whether the function declaration says `const *FuriString` or `*FuriString`.

mogenson:
After inspecting the flipperzero-firmware code, `dialog_show_file_browser()` will not modify the preselected file `path` FuriString, but some common storage processing functions will replace `/ext`, `/app`, `/any` prefixes depending on the context. I think `pub(crate) fn as_mut_ptr(&mut self) -> *mut sys::FuriString` make sense, at least for the `show_file_browser()` usage. Is everyone in agreement?

mogenson:
Also, is there anything I need to do with this PR to get the Clippy (MSRV) lints CI to succeed?

mogenson:
@dcoles @JarvisCraft @str4d Are there any remaining issues or is this PR able to be merged?

JarvisCraft (approved):

dcoles:
> Also, is there anything I need to do with this PR to get the Clippy (MSRV) lints CI to succeed?

@str4d Would you be able to take a look at this? I'm not quite sure why the beta is fine, but MSRV is throwing a `Unable to run @augu/clippy-action: HttpError: Resource not accessible by integration`.

dcoles:
> @dcoles @JarvisCraft @str4d Are there any remaining issues or is this PR able to be merged?

Yes, let's get this merged. Thanks very much for your patience and for the PR!

mogenson:
Cheers, thanks!

PR #81, opened 2023-06-08
docs: enhance documentation support and resolve clippy warnings

JarvisCraft:
# Description

This enhances documentation support by:

* resolving existing docs issues:
  * invalid links;
  * missing `# Safety` sections
* enabling automatic generation of `# Features` section;
* resolving `rustdoc` and `clippy` warnings which were polluting the log.

### DIFF
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -24,7 +24,7 @@ flipperzero-rt = { path = "rt", version = "0.10.0" }
 flipperzero-alloc = { path = "alloc", version = "0.10.0" }
 flipperzero-test = { path = "test", version = "0.10.0" }
 ufmt = "0.2.0"
-document-features ="0.2.0"
+document-features = "0.2.0"
 
 [profile.dev]
 opt-level = "z"
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -24,6 +24,7 @@ flipperzero-rt = { path = "rt", version = "0.10.0" }
 flipperzero-alloc = { path = "alloc", version = "0.10.0" }
 flipperzero-test = { path = "test", version = "0.10.0" }
 ufmt = "0.2.0"
+document-features = { version = "0.2.0" }
 
 [profile.dev]
 opt-level = "z"
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -24,7 +24,7 @@ flipperzero-rt = { path = "rt", version = "0.10.0" }
 flipperzero-alloc = { path = "alloc", version = "0.10.0" }
 flipperzero-test = { path = "test", version = "0.10.0" }
 ufmt = "0.2.0"
-document-features = { version = "0.2.0" }
+document-features ="0.2.0"
 
 [profile.dev]
 opt-level = "z"
--- a/crates/flipperzero/Cargo.toml
+++ b/crates/flipperzero/Cargo.toml
@@ -35,6 +35,9 @@ lock_api = "0.4"
 digest = "0.10"
 bitflags = "1.0"
 
+# Docs
+document-features = { workspace = true, optional = true }
+
 [dev-dependencies]
 flipperzero-alloc.workspace = true
 flipperzero-rt.workspace = true
@@ -45,8 +48,24 @@ md-5 = { version = "0.10", default-features = false }
 sha2 = { version = "0.10", default-features = false }
 
 [features]
-# enables features requiring an allocator
+
+#! ## Core features
+
+## Enables features requiring an allocator.
+##
+## This will require you to set up an allocator such as by the following:
+##
+## ```
+## extern crate alloc;
+## extern crate flipperzero_alloc;
+## ```
 alloc = []
+
+# Internal features.
+
+# Enables unstable documentation features.
+# This is only intended to be used by `cargo doc`.
+unstable_docs = []
 
 [[test]]
 name = "dolphin"
--- a/crates/flipperzero/Cargo.toml
+++ b/crates/flipperzero/Cargo.toml
@@ -53,7 +53,8 @@ sha2 = { version = "0.10", default-features = false }
 
 ## Enables features requiring an allocator.
 ##
-## This will require you to set up an allocator such as by the following:
+## This will require you to set the global allocator
+## by importing the `flipperzero_alloc`:
 ##
 ## ```
 ## extern crate alloc;
--- a/crates/flipperzero/Cargo.toml
+++ b/crates/flipperzero/Cargo.toml
@@ -61,12 +61,6 @@ sha2 = { version = "0.10", default-features = false }
 ## ```
 alloc = []
 
-# Internal features.
-
-# Enables unstable documentation features.
-# This is only intended to be used by `cargo doc`.
-unstable_docs = []
-
 [[test]]
 name = "dolphin"
 harness = false
--- a/crates/flipperzero/src/furi/string.rs
+++ b/crates/flipperzero/src/furi/string.rs
@@ -35,13 +35,16 @@ const WHITESPACE: &[char] = &[
 ///
 /// This is similar to Rust's [`CString`] in that it represents an owned, C-compatible,
 /// nul-terminated string with no nul bytes in the middle. It also has additional methods
-/// to provide the flexibility of Rust's [`String`](alloc::string::String).
+/// to provide the flexibility of Rust's [`String`].
 /// It is used in various APIs of the Flipper Zero SDK.
 ///
 /// This type does not requre the `alloc` feature flag, because it does not use the Rust
 /// allocator. Very short strings (7 bytes or fewer) are stored directly inside the
 /// `FuriString` struct (which is stored on the heap), while longer strings are allocated
 /// on the heap by the Flipper Zero firmware.
+///
+/// [`CString`]: alloc::ffi::CString
+/// [`String`]: alloc::string::String
 #[derive(Eq)]
 pub struct FuriString(NonNull<sys::FuriString>);
 
--- a/crates/flipperzero/src/furi/string.rs
+++ b/crates/flipperzero/src/furi/string.rs
@@ -35,8 +35,8 @@ const WHITESPACE: &[char] = &[
 ///
 /// This is similar to Rust's [`CString`] in that it represents an owned, C-compatible,
 /// nul-terminated string with no nul bytes in the middle. It also has additional methods
-/// to provide the flexibility of Rust's [`String`]. It is used in various APIs of the
-/// Flipper Zero SDK.
+/// to provide the flexibility of Rust's [`String`](alloc::string::String).
+/// It is used in various APIs of the Flipper Zero SDK.
 ///
 /// This type does not requre the `alloc` feature flag, because it does not use the Rust
 /// allocator. Very short strings (7 bytes or fewer) are stored directly inside the
@@ -292,70 +292,61 @@ impl FuriString {
         Bytes(self.to_bytes().iter().copied())
     }
 
-    /// Returns `true` if the given pattern matches a sub-slice of this string slice.
+    /// Returns `true` if the given pattern `pat` matches a sub-slice of this string slice.
     ///
     /// Returns `false` if it does not.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`],
+    /// or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     #[inline]
     pub fn contains<P: Pattern>(&self, pat: P) -> bool {
         pat.is_contained_in(self)
     }
 
-    /// Returns `true` if the given pattern matches a prefix of this string slice.
+    /// Returns `true` if the given pattern `pat` matches a prefix of this string slice.
     ///
     /// Returns `false` if it does not.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     pub fn starts_with<P: Pattern>(&self, pat: P) -> bool {
         pat.is_prefix_of(self)
     }
 
-    /// Returns `true` if the given pattern matches a suffix of this string slice.
+    /// Returns `true` if the given pattern `pat` matches a suffix of this string slice.
     ///
     /// Returns `false` if it does not.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     pub fn ends_with<P: Pattern>(&self, pat: P) -> bool {
         pat.is_suffix_of(self)
     }
 
-    /// Returns the byte index of the first byte of this string that matches the pattern.
+    /// Returns the byte index of the first byte of this string that matches the pattern `pat`.
     ///
     /// Returns [`None`] if the pattern doesn't match.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     #[inline]
     pub fn find<P: Pattern>(&self, pat: P) -> Option<usize> {
         pat.find_in(self)
     }
 
-    /// Returns the byte index for the first byte of the last match of the pattern in this
-    /// string.
+    /// Returns the byte index for the first byte of the last match of the pattern `pat`
+    /// in this string.
     ///
     /// Returns [`None`] if the pattern doesn't match.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     #[inline]
     pub fn rfind<P: Pattern>(&self, pat: P) -> Option<usize> {
         pat.rfind_in(self)
@@ -404,23 +395,19 @@ impl FuriString {
 
     /// Repeatedly removes from this string all prefixes and suffixes that match a pattern.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`] or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     pub fn trim_matches<P: Pattern + Copy>(&mut self, pat: P) {
         self.trim_start_matches(pat);
         self.trim_end_matches(pat);
     }
 
-    /// Repeatedly removes from this string all prefixes that match a pattern.
+    /// Repeatedly removes from this string all prefixes that match a pattern `pat`.
     ///
-    /// The [pattern] can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`] or a slice of [`char`]s.
     ///
     /// [`char`]: prim@char
-    /// [pattern]: self::pattern
     ///
     /// # Text directionality
     ///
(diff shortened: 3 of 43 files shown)


JarvisCraft:
It;s worth noting that

```toml
[workspace.package.metadata.docs.rs]
all-features = true
```

in workspace root failed to work (this section seems to be [ignored](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-metadata-table), at least for this specific field).

JarvisCraft:
@str4d, I've merged the latest `main` so that it now should comply with #83.

crates/flipperzero/src/furi/string.rs
```diff
@@ -35,8 +35,8 @@ const WHITESPACE: &[char] = &[
 ///
 /// This is similar to Rust's [`CString`] in that it represents an owned, C-compatible,
 /// nul-terminated string with no nul bytes in the middle. It also has additional methods
-/// to provide the flexibility of Rust's [`String`]. It is used in various APIs of the
-/// Flipper Zero SDK.
+/// to provide the flexibility of Rust's [`String`].
+/// It is used in various APIs of the Flipper Zero SDK.
```
str4d, line 39: This change is no longer necessary:
```suggestion
/// to provide the flexibility of Rust's [`String`]. It is used in various APIs of the
/// Flipper Zero SDK.
```

crates/flipperzero/src/lib.rs
```diff
@@ -1,10 +1,14 @@
 //! High-level bindings for the Flipper Zero.
-
```
str4d, line 2: Please leave this space in (between the documentation and `#![no_std]` line). Note that `//!` is syntactic sugar for `#![doc]`, so the latter should not be treated as part of the module-level block of attributes.

crates/flipperzero/src/lib.rs
```diff
@@ -1,10 +1,14 @@
...
+//!
+//! # Features
+#![cfg_attr(feature = "document-features", doc = document_features::document_features!())]
 #![no_std]
 #![cfg_attr(test, no_main)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![deny(rustdoc::broken_intra_doc_links)]
 
-#[cfg(any(feature = "alloc", docsrs))]
+#[cfg(feature = "alloc")]
```
str4d, line 10: This change breaks intra-doc links. Revert it.
```suggestion
#[cfg(any(feature = "alloc", docsrs))]
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -1,13 +1,19 @@
 //! Inlines for Furi HAL GPIO interface.
 //!
-//! See: <https://github.com/flipperdevices/flipperzero-firmware/blob/release/firmware/targets/f7/furi_hal/furi_hal_gpio.h>
+//! See: [furi_hal_gpio.h][1]
```
str4d, line 3:
```suggestion
//! See: [`furi_hal_gpio.h`][1]
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -1,13 +1,19 @@
...
 use crate as sys;
 
 /// Number of GPIO on one port.
 pub const GPIO_NUMBER: usize = 16;
 
 /// GPIO write pin.
+///
+/// # Safety
+///
+/// Safety invariants are not yet defined for GPIO.
```
str4d, line 16:
```suggestion
/// `gpio` must be non-null, and the memory it points to must be initialized.
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -17,6 +23,10 @@ pub unsafe extern "C" fn furi_hal_gpio_write(gpio: *const sys::GpioPin, state: b
 }
 
 /// GPIO write pin.
+///
+/// # Safety
+///
+/// Safety invariants are not yet defined for GPIO.
```
str4d, line 29:
```suggestion
/// `port` must be non-null, and the memory it points to must be initialized.
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -28,6 +38,10 @@ pub unsafe extern "C" fn furi_hal_gpio_write_port_pin(
 }
 
 /// GPIO read pin.
+///
+/// # Safety
+///
+/// Safety invariants are not yet defined for GPIO.
```
str4d, line 44:
```suggestion
/// `gpio` must be non-null, and the memory it points to must be initialized.
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -37,6 +51,10 @@ pub unsafe extern "C" fn furi_hal_gpio_read(gpio: *const sys::GpioPin) -> bool {
...
```
str4d, line 57:
```suggestion
/// `port` must be non-null, and the memory it points to must be initialized.
```

crates/sys/src/lib.rs
```diff
@@ -22,8 +22,9 @@ mod inlines;
     non_upper_case_globals,
     non_camel_case_types,
     non_snake_case,
-    clippy::transmute_int_to_bool,
     clippy::useless_transmute,
+    clippy::transmute_int_to_bool,
+    clippy::missing_safety_doc,
```
str4d, line 27: Please minimise the diff here:
```suggestion
    clippy::missing_safety_doc,
    clippy::transmute_int_to_bool,
    clippy::useless_transmute,
```

tools/src/bin/storage.rs
```diff
@@ -1,6 +1,8 @@
 //! Storage CLI.
 //!
-//! See: <https://github.com/flipperdevices/flipperzero-firmware/blob/dev/scripts/storage.py>
+//! See: [storage.py][1] script.
```
str4d, line 3:
```suggestion
//! See: [`storage.py`][1] script.
```

tools/src/bin/storage.rs
```diff
@@ -1,6 +1,8 @@
...
+//!
+//! [1]: (https://github.com/flipperdevices/flipperzero-firmware/blob/dev/scripts/storage.py)
```
str4d, line 5:
```suggestion
//! [1]: https://github.com/flipperdevices/flipperzero-firmware/blob/dev/scripts/storage.py
```

crates/flipperzero/Cargo.toml
```diff
@@ -45,7 +48,17 @@ md-5 = { version = "0.10", default-features = false }
 sha2 = { version = "0.10", default-features = false }
 
 [features]
-# enables features requiring an allocator
+
+#! ## Core features
+
+## Enables features requiring an allocator.
+##
+## This will require you to set up an allocator such as by the following:
```
str4d, line 56:
```suggestion
## This will require you to set the global allocator by importing the `flipperzero_alloc`
## crate:
```

crates/Cargo.toml
```diff
@@ -24,6 +24,7 @@ flipperzero-rt = { path = "rt", version = "0.10.0" }
 flipperzero-alloc = { path = "alloc", version = "0.10.0" }
 flipperzero-test = { path = "test", version = "0.10.0" }
 ufmt = "0.2.0"
+document-features = { version = "0.2.0" }
```
str4d, line 27:
```suggestion
document-features = "0.2.0"
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -295,63 +295,59 @@ impl FuriString {
...
 
-    /// Returns `true` if the given pattern matches a sub-slice of this string slice.
+    /// Returns `true` if the given pattern `pat` matches a sub-slice of this string slice.
     ///
     /// Returns `false` if it does not.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`],
+    /// or a slice of [`char`]s.
```
str4d, line 303: Undo this change; the line wrap length was correct for this file (90 characters).
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -295,63 +295,59 @@ impl FuriString {
...
     }
 
-    /// Returns `true` if the given pattern matches a prefix of this string slice.
+    /// Returns `true` if the given pattern `pat` matches a prefix of this string slice.
     ///
     /// Returns `false` if it does not.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
```
str4d, line 315: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -295,63 +295,59 @@ impl FuriString {
...
     }
 
-    /// Returns `true` if the given pattern matches a suffix of this string slice.
+    /// Returns `true` if the given pattern `pat` matches a suffix of this string slice.
     ///
     /// Returns `false` if it does not.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
```
str4d, line 326: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -295,63 +295,59 @@ impl FuriString {
...
     }
 
-    /// Returns the byte index of the first byte of this string that matches the pattern.
+    /// Returns the byte index of the first byte of this string that matches the pattern `pat`.
     ///
     /// Returns [`None`] if the pattern doesn't match.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
```
str4d, line 337: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -295,63 +295,59 @@ impl FuriString {
...
     ///
     /// [`char`]: prim@char
     pub fn ends_with<P: Pattern>(&self, pat: P) -> bool {
         pat.is_suffix_of(self)
     }
 
-    /// Returns the byte index of the first byte of this string that matches the pattern.
+    /// Returns the byte index of the first byte of this string that matches the pattern `pat`.
```
str4d, line 333:
```suggestion
    /// Returns the byte index of the first byte of this string that matches the pattern
    /// `pat`.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -295,63 +295,59 @@ impl FuriString {
...
-    /// Returns the byte index for the first byte of the last match of the pattern in this
-    /// string.
+    /// Returns the byte index for the first byte of the last match of the pattern `pat`
+    /// in this string.
     ///
     /// Returns [`None`] if the pattern doesn't match.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
```
str4d, line 350: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -402,19 +398,17 @@ impl FuriString {
 
     /// Repeatedly removes from this string all prefixes and suffixes that match a pattern.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`] or a slice of [`char`]s.
```
str4d, line 401: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -402,19 +398,17 @@ impl FuriString {
...
         self.trim_start_matches(pat);
         self.trim_end_matches(pat);
     }
 
-    /// Repeatedly removes from this string all prefixes that match a pattern.
+    /// Repeatedly removes from this string all prefixes that match a pattern `pat`.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`] or a slice of [`char`]s.
```
str4d, line 411: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -428,10 +422,9 @@ impl FuriString {
         while self.strip_prefix(pat) {}
     }
 
-    /// Repeatedly removes from this string all suffixes that match a pattern.
+    /// Repeatedly removes from this string all suffixes that match a pattern `pat`.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`] or a slice of [`char`]s.
```
str4d, line 427: Ditto:
```suggestion
    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -445,31 +438,29 @@ impl FuriString {
...
+    /// Removes the given `prefix` from this string.
     ///
     /// If the string starts with the pattern `prefix`, returns `true`. Unlike
     /// [`Self::trim_start_matches`], this method removes the prefix exactly once.
     ///
     /// If the string does not start with `prefix`, returns `false`.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The prefix can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
```
str4d, line 448:
```suggestion
    /// The prefix can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string.rs
```diff
@@ -445,31 +438,29 @@ impl FuriString {
...
+    /// Removes the given `suffix` from this string.
     ///
     /// If the string ends with the pattern `suffix`, returns `true`. Unlike
     /// [`Self::trim_end_matches`], this method removes the suffix exactly once.
     ///
     /// If the string does not end with `suffix`, returns `false`.
     ///
-    /// The pattern can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
-    /// [`char`]s.
+    /// The suffix can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of [`char`]s.
```
str4d, line 463:
```suggestion
    /// The suffix can be a `&FuriString`, [`c_char`], `&CStr`, [`char`], or a slice of
    /// [`char`]s.
```

crates/flipperzero/src/furi/string/pattern.rs
```diff
@@ -36,14 +36,17 @@ const FURI_STRING_FAILURE: usize = usize::MAX;
 pub trait Pattern: Sized {
     /// Checks whether the pattern matches anywhere in the haystack.
     #[inline]
+    #[allow(clippy::wrong_self_convention)] // same as in `std`
```
str4d, line 39: If these are being moved onto the trait methods, then remove the blanket allow on the trait that I added in #83.

str4d (changes requested):
Generally LGTM

crates/flipperzero/src/furi/string.rs
JarvisCraft: I think that it is more readable now (when looking at docs in sources) since the new sentence starts from the new line while still maintaining approximately same column width.

crates/flipperzero/src/lib.rs
JarvisCraft: I also think that the space fits better, but for some reasong `cargo fmt` prefers to remove the blank line.

crates/flipperzero/src/lib.rs
JarvisCraft: I think that it would be better to enable `--all-features` for docs, but I will bring this back as for mow.

crates/flipperzero/src/furi/string/pattern.rs
JarvisCraft: Yup, thanks. Didn't see the top-level attribute when resolving the conflicts

crates/flipperzero/src/lib.rs
JarvisCraft: Done

crates/flipperzero/src/furi/string.rs
JarvisCraft: Afaik, the [default](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#doc_comment_code_block_width) for docomment width is also `100` so we don't even need tro split the lines.

Although if you insist, we could enforce `90` though it does not seems reasonable to me (and, in this speific case, 100 looks better IMO).

Same goes for the other similar comments although you ma

crates/flipperzero/src/lib.rs
str4d: Hmm. What if you add a `//!` before the space? That should be valid documentation, and should prevent rustfmt from joining the two.

JarvisCraft: Done

crates/flipperzero/src/furi/string.rs
str4d: That default is for code snippets in doc comments, not for documentation itself. I presume it's set to 100 because that is the same break point as for regular code.

I use 90 characters as a break point for documentation because it hits a nice spot for IDE and terminal development readability. 100 for code works well because it's rare you have lines that long, and when the formatter does trigger a break the resulting lines end up much shorter. For documentation, that is not the case; the break point ends up forming a block boundary as there are more granular points to break at, so you end up with mostly full-width lines, and for larger paragraphs 100 is too wide for easy reading. The research done on website UX tends to show that readability drops off at line lengths above 80-85 characters, so including the 4-8 character prefixes on doc lines, 90 sits at a nice break point for readability.

But in any case, what I care about more is minimizing PR diffs. Every change consumes reviewer energy, while they determine if it is semantically meaningful or not. If less energy is spent verifying that a change is just altering formatting, more energy is available for noticing bugs. So I prefer formatting-only changes to be separated out into separate PRs or commits, and ideally automated (though sadly AFAIK `rustfmt` doesn't have even optional support for formatting documentation).

JarvisCraft: A very good point! Thanks for the detailed explanation.

In this case, I will revert the changes to this lines.

Also, let's add this experimental option to the formatting file in the future so that it is more explicit for other committers.

JarvisCraft:
@str4d, I guess there are no more changes here and the CI seems to fail due to some external error

str4d:
Yeah, I've been trying to figure out a way around it, but unfortunately (for good security reasons) GitHub doesn't allow workflows run on PRs from third-party repos to have _any_ write permissions to the base repo. And the Clippy action I found does a small write at the start to create a status entry if it doesn't exist, and _that_ is what is blocking the actual Clippy checks from running. I've opened https://github.com/auguwu/clippy-action/issues/15 upstream for this, but if this can't be addressed easily there, then I'll look into an alternative approach.

crates/Cargo.toml
```diff
@@ -24,6 +24,7 @@ flipperzero-rt = { path = "rt", version = "0.10.0" }
 flipperzero-alloc = { path = "alloc", version = "0.10.0" }
 flipperzero-test = { path = "test", version = "0.10.0" }
 ufmt = "0.2.0"
+document-features ="0.2.0"
```
str4d, line 27:
```suggestion
document-features = "0.2.0"
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -37,6 +53,12 @@ pub unsafe extern "C" fn furi_hal_gpio_read(gpio: *const sys::GpioPin) -> bool {
...
+///
```
str4d, line 60: We've defined the safety invariants we require for this Rust code. Any other requirements relate to the correctness of the inline method's API as called by the SDK.
```suggestion
```

crates/sys/src/inlines/furi_hal_gpio.rs
```diff
@@ -17,6 +23,12 @@ pub unsafe extern "C" fn furi_hal_gpio_write(gpio: *const sys::GpioPin, state: b
 }
 
 /// GPIO write pin.
+///
+/// # Safety
+///
+/// Safety invariants are not yet defined for GPIO.
+///
```
str4d, line 30:
```suggestion
```

str4d (approved):
ACK with a few cleanups.

PR #82, opened 2023-06-08
ci: use caching actions to setup build environment

JarvisCraft:
# Description

This enables the usage of GitHub Actions which cache dependencies by default
in order to speed up builds.

### DIFF
--- a/.github/workflows/_build_and_test.yml
+++ b/.github/workflows/_build_and_test.yml
@@ -14,10 +14,14 @@ jobs:
       run:
         working-directory: ${{ inputs.target }}
     steps:
-    - uses: actions/checkout@v3
-    - run: sudo apt install libudev-dev
-    - name: Rustup
-      run: rustup +nightly target add thumbv7em-none-eabihf
+    - uses: actions/checkout@v3.5.2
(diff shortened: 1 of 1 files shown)


JarvisCraft:
`@dtolnay`'s action seems to not perform caching by itself and `@actions-rs` are unmaintained. Probably manual caching should be done for the former.

dcoles:
It looks like running `apt install` isn't too bad in our existing GitHub actions (I've seen it range from [4s](https://github.com/flipperzero-rs/flipperzero/actions/runs/5162160023/jobs/9299661889) to [16s](https://github.com/flipperzero-rs/flipperzero/actions/runs/5212435375/jobs/9406026313)).

Though it definately seems a good idea to use `@dtolnay`'s toolchain action.

str4d:
#83 now relies on `rust-toolchain.toml` as the baseline source of truth, and uses the `toolchain` action where needed (to switch to e.g. beta).

JarvisCraft:
Could this PR be merged at its current state so that I can actualize #83?

JarvisCraft:
Superseded by #83

Issue #58, opened 2023-03-18
Support stable Rust toolchain


dcoles:
Currently Rust for Flipper Zero [requires a nightly toolchain](https://github.com/flipperzero-rs/flipperzero/blob/19b185e63f60572da8a7418185f5e5b87f6f6959/crates/rust-toolchain.toml) to build binaries. This makes development somewhat more difficult as we have to regularly update this to pull in new features or bug fixes.

## Blockers to supporting a stable Rust toolchain

- [x] rust-lang/rust/issues/66741: [Stable as of Rust 1.68.0](https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#default-alloc-error-handler)
- [ ] rust-lang/rust/issues/90275: Flipper Zero's dynamic linker does not support unique section names

dcoles:
A possible alternative to waiting for [`no-unique-section-names`](rust-lang/rust/issues/90275) is to teach the Flipper Zero linker about non-unique section names.

dcoles:
It appears that the Flipper Zero linker can now handle non-unique section names making `-Z no-unique-section-names` uneeded.

PR #83, opened 2023-06-08
CI updates

str4d:


.github/workflows/_lints.yml
```diff
@@ -0,0 +1,67 @@
...
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    defaults:
+      run:
+        working-directory: ${{ inputs.target }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@beta
+        id: toolchain
+      - run: rustup override set ${{steps.toolchain.outputs.name}}
```
JarvisCraft:
```suggestion
      - run: rustup override set ${{ steps.toolchain.outputs.name }}
```

.github/workflows/_lints.yml
```diff
@@ -0,0 +1,77 @@
...
+
+jobs:
+  clippy:
+    name: Clippy (MSRV)
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ inputs.target }}
+    steps:
+      - uses: actions/checkout@v3
```
JarvisCraft, line 18: I suggest that exact versions are used where possible such as

```suggestion
      - uses: actions/checkout@v3.5.2
```

and Dependanabot is configured to update versions as done [here](https://github.com/JarvisCraft/mc-server-pinger/blob/4d9248575e24a6c914e612ba3083d6f89ec2887d/.github/dependabot.yml#L25-L34) (see PR history in the repo to see how it looks).

It will allow better reproducibility while keeping up to date.

.github/workflows/_lints.yml
JarvisCraft: *PS*: For some reason GH dies not render `Resolve thread` button for me

.github/workflows/_lints.yml
str4d: Yep, Dependabot config is coming next once I get the lints working.

.github/workflows/_lints.yml
```diff
@@ -0,0 +1,83 @@
...
+            --manifest-path ${{ inputs.target }}/Cargo.toml
+            --all-features
+            --all-targets
+            --
+            -D warnings
+
+  clippy-beta:
+    name: Clippy (beta)
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
str4d, line 36: I'm adding this like I do in other repos I maintain, because it's useful for checking whether there are Clippy lint changes coming down the pipeline that a future MSRV bump may encounter, without the noise of checking lints against nightly.

However, until #58 is closed, this step will silently fail, and the `Clippy (MSRV)` step _will_ be checking lints against nightly. So this step won't be useful until then, but it also won't get in the way, so we can add it now in preparation.

.github/workflows/_lints.yml
```diff
@@ -0,0 +1,83 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ inputs.target }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install build dependencies
+        run: sudo apt install libudev-dev
+      - name: Run Clippy
+        uses: actions-rs/clippy-check@v1
```
JarvisCraft, line 22: If `actions-rs` is used anyway than maybe it is OK to use their `toolchain` action too.

My motivation to not use it in #82 was that this actions are not maintained and use some deprecated GH features (such as Node 12), but if at least one of them has to be used (`clippy-check`) then there is no good readon to omit the other ones.

JarvisCraft (changes requested):

.github/workflows/_lints.yml
str4d: I kept using this one action because I had not yet found a viable alternative, but I just did: https://github.com/auguwu/clippy-action

str4d:
The two remaining CI failures are:
- Clippy lints about missing documentation, that will be resolved by #81.
- Intra-doc link checking failure for `tools` due to it using `stable` and my having just added `--cfg docsrs` to CI (so that I can enable `alloc` for automated link resolving).
  - I will probably just change this lint to always use nightly, since that's what docs.rs does.

.github/workflows/_lints.yml
dcoles: It looks like the `auguwu/clippy-action` action will add source-code annotations when run. Does this mean we'll get duplicate annotations?

crates/sys/src/lib.rs
```diff
@@ -20,6 +21,9 @@ mod inlines;
 #[allow(non_upper_case_globals)]
 #[allow(non_camel_case_types)]
 #[allow(non_snake_case)]
+#[allow(clippy::transmute_int_to_bool)]
+#[allow(clippy::useless_transmute)]
```
dcoles, line 25: It might make sense to `#[allow(clippy::all)]`, since it's unlikely there's much we can do about any warnings in bindgen code.

I think the equivalent for rustc lints is `#[allow(warnings, unused)]`.

.github/workflows/_lints.yml
```diff
@@ -0,0 +1,77 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install build dependencies
+        run: sudo apt install libudev-dev
+      - name: Run Clippy
+        uses: auguwu/clippy-action@1.1.2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          working-directory: ${{ inputs.target }}
+          args: --all-features --all-targets -- -D warnings
```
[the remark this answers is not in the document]
str4d: This check is specifically against MSRV as pinned in `rust-toolchain.toml`. So the only time new warnings would appear is if your PR is changing that pinned toolchain, and then it's not unrelated to your PR. Once we get onto stable Rust, this will only occur in PRs that bump MSRV.

.github/workflows/_lints.yml
str4d: For `actions-rs/clippy-check` this wasn't the case with `continue-on-error: true` AFAIR, but IDK about this action. At least from what I've seen in this PR itself, we weren't getting duplicates.

crates/sys/src/lib.rs
str4d: Eh, what we have here is sufficient for closing the current warnings, and if something new comes up it is probably worth us noticing so we can open an issue on `bindgen`.

JarvisCraft: By the way, would not

```suggestion
#[allow(
    non_upper_case_globals,
    non_camel_case_types,
    non_snake_case,
    clippy::transmute_int_to_bool
)]
```

look more compact here?

.github/workflows/_lints.yml
```diff
@@ -0,0 +1,77 @@
...
+    defaults:
+      run:
+        working-directory: ${{ inputs.target }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install build dependencies
+        run: sudo apt install libudev-dev
+      - name: Run Clippy
+        uses: auguwu/clippy-action@1.1.2
+        with:
```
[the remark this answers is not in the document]
str4d: The `deny` etc. options are currently broken in the action. I've opened an issue upstream: https://github.com/auguwu/clippy-action/issues/13

The reason my workflow worked before is because `args` was being completely ignored. I've opened an issue for that upstream as well: https://github.com/auguwu/clippy-action/issues/14

So currently there is no way to configure anything except `--all-features`. I'll comment out the `deny` lines for now, and leave `--all-targets` in the ignored `args` input, and then once an update to the action is available, we can re-enable these.

crates/sys/src/lib.rs
str4d: Indeed. Done.

.github/workflows/_lints.yml
str4d: Dependabot config has been added.

str4d:
Okay, I've addressed all comments, and the only failing lint is missing documentation that will be added by #81.

JarvisCraft:
Could this PR be merged at its current state so that I can actualize #81 with the required fixes?

str4d:
Yes, I think so.

PR #90, opened 2023-06-26
Fix gpio read always returning true

agarof:
Also make register manipulations volatile as the original `furi` hal functions are.
Updated the GPIO example to verify that the actual state was being returned.


dcoles:
Hi @agarof. Thanks for the PR!

Will merge that in right now.

PR #91, opened 2023-07-05
Fix cargo doc when not using the alloc feature

agarof:
`cargo doc` could not build the documentation when the `alloc` feature was disabled because it could not access `alloc` symbols.
This replaces them with direct links to the nightly documentation.


JarvisCraft:
Imo, the docs are intended to be built with `--all-features`

dcoles:
Hi @agarof,

Thanks for the PR! This is a bit of a weird case, since it's referring to a feature that might not be supported (which is part of the purpose of having `FuriString`). This seems to be a good compromise.

PR #92, opened 2023-07-08
feat: make `DialogFileBrowserOptions::new()` use native initialization function

JarvisCraft:
# Description

In #80 safe bindings for dialogs app file browser were added by @mogenson. At this moment, the native initialization function was unsound since it forgot to initialize one of the fields so we decided to perform initialization on Rust side.

Now that this issues is fixed in flipperdevices/flipperzero-firmware#2756 with the future guarantee to always initialize all fields we can (and should) migrate to this approach which is better for us from backwards compatibility standpoint since we will no longer have to reflect the changes to this structure on our side for sake of safety.

# Requirements

- [x] This PR requires the bindings to be updated to firmware version `0.86.1` which contains the required fix to the `dialog_file_browser_set_basic_options`.

# Notes

I've also added a `c_str` macro which has magically become `const`-friendly with the recent stabilization of `from_bytes_with_nul` which will land to stable soon (at least, by the time we are ready to become stable-friendly; anyways, it is behind  a macro so there is no explicit dependency on unstable toolchain as for now).

### DIFF
--- a/crates/flipperzero/src/dialogs/mod.rs
+++ b/crates/flipperzero/src/dialogs/mod.rs
@@ -5,6 +5,7 @@ use alloc::ffi::CString;
 
 use core::ffi::{c_void, CStr};
 use core::marker::PhantomData;
+use core::mem::MaybeUninit;
 use core::ptr::{self, NonNull};
 
 use flipperzero_sys as sys;
@@ -25,6 +26,7 @@ pub struct DialogMessage<'a> {
 }
 
 /// A dialog file browser options.
+#[repr(transparent)]
 pub struct DialogFileBrowserOptions<'a> {
     data: sys::DialogsFileBrowserOptions,
     _phantom: PhantomData<&'a ()>,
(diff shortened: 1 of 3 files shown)


mogenson (approved):
Lgtm!

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -202,19 +203,22 @@ impl DialogMessageButton {
 
 impl<'a> DialogFileBrowserOptions<'a> {
     /// Creates a new dialog file browser options and initializes to default values.
-    pub fn new() -> Self {
+    pub fn new<'e: 'a>(extension: &'e CStr) -> Self {
```
dcoles, line 206: I'm not familiar with the `'e: 'a` syntax (I think it's saying `'e` must live for at least as long as `'a`).

What would the difference be between `'e: 'a` and `'a` on its own?

crates/sys/src/lib.rs
```diff
@@ -29,12 +29,24 @@ mod inlines;
...
+        ::core::concat!($str, "\0").as_ptr() as *const ::core::ffi::c_char
+    }};
+}
+
+/// Create a static C string of type [`&CStr`][`core::ffi::CStr`].
+/// Will automatically add a NUL terminator.
+#[macro_export]
+macro_rules! c_str {
+    ($str:expr $(,)?) => {{
+        match ::core::ffi::CStr::from_bytes_with_nul(::core::concat!($str, "\0").as_bytes()) {
```
dcoles, line 46: Since this is guaranteed to return a string with a NUL terminator, then we should be able to use [`CStr::from_bytes_with_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked).

(I'm not 100% sure how macros behave with `#![deny(unsafe_code)]`. It may be worth switching to a generic const function if it does trigger the lint.)

crates/flipperzero/src/dialogs/mod.rs
JarvisCraft: Exactly, this is `'e outlives 'a`. In my understanding, if both lifetimes were the same, we would not be able to have `&'static extensions` without `'a` also being `'static`, although I may be wrong here

crates/sys/src/lib.rs
JarvisCraft: When used in `const` code (which is the expected use-case for the macro), panics should not end in the resulting code since the compiler should evaluate the resulting value, so I don't think that `panic` gere iş A problem.

As for `unsafe` in macro expansions, I am not sure how it interacts with `#![forbid(unsafe)] `, although proc-macro [`cstr`](https://docs.rs/cstr/0.2.11/cstr/) does generate `unsafe` call for this use-case.

By the way, it may be more rational to just re-export this macro from `cstr` crate rather than ad-hoc implement it here (specially, considering the fact that it does intermediate null-checking).

dcoles: Yeah. I'm not at all opposed to pulling in a crate if it provides better usability.

Currently the other benefit of [`CStr::from_bytes_with_nul_unchecked`]([CStr::from_bytes_with_nul](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked) is that it has been `const` since 1.59.0.

Does `expect` work correctly in `const` evaluation? If so, that would be a better choice than `match`/`panic`.

crates/flipperzero/src/dialogs/mod.rs
dcoles: I think this is where [variance](https://doc.rust-lang.org/reference/subtyping.html) comes into play.

Consider the [case of a `struct` with `&'a str`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5c33d5bcd5cc1b51a10c2606a005884e). Without variance we couldn't use `&'static str`  without also making the struct `'static`.

So I think in this case we would only need `'a`.

crates/sys/src/lib.rs
JarvisCraft: I've replaced the use with `cstr` crate which uses the `_unchecked` variant internally.
I've also made `c_string` delegate to it which has a minor memory overhead (since it now may statically store `CStr`'s meta) but it seems appropriate since we now get better compile-time checking.

In the future we may want to replace this with our own solution (or finally enjoy `c"foo"` syntax), but as for now it looks like the best solution.

crates/flipperzero/src/dialogs/mod.rs
JarvisCraft: Fair enough

I've replaced this with `'a` and added two doc-tests to validate appropriate and inappropriate use-cases.

crates/sys/src/lib.rs
```diff
@@ -29,12 +29,18 @@ mod inlines;
 )]
 mod bindings;
 
-/// Create a static C string.
+/// Create a static C string of type [`&'static CStr`][`core::ffi::CStr`].
 /// Will automatically add a NUL terminator.
+pub use cstr::cstr;
```
str4d, line 34: Note this addition in the changelog under the "Added" section:
```markdown
- `flipperzero_sys::cstr`
```

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -25,6 +25,7 @@ pub struct DialogMessage<'a> {
 }
 
 /// A dialog file browser options.
+#[repr(transparent)]
 pub struct DialogFileBrowserOptions<'a> {
```
str4d, line 29: I note that this new struct was not documented as an addition in the changelog. As we haven't cut a release with it yet, we don't need to note the changes in this PR; just add the entire struct as an addition.

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -5,10 +5,10 @@ use alloc::ffi::CString;
 
 use core::ffi::{c_char, c_void, CStr};
 use core::marker::PhantomData;
+use core::mem::MaybeUninit;
 use core::ptr::{self, NonNull};
-
```
str4d, line 9: Undo the deletion of this newline (it is intentionally separating the Rust core changes from the external dependencies).

crates/flipperzero/src/dialogs/mod.rs
```diff
@@ -25,6 +25,7 @@ pub struct DialogMessage<'a> {
 }
 
 /// A dialog file browser options.
+#[repr(transparent)]
```
str4d, line 28: Why is this necessary?

crates/flipperzero/src/dialogs/mod.rs
JarvisCraft: Reverted

crates/sys/src/lib.rs
JarvisCraft: Done

crates/flipperzero/src/dialogs/mod.rs
JarvisCraft: Good point, added it

crates/flipperzero/src/dialogs/mod.rs
JarvisCraft: While it is not used explicitly here, it allows tricks like pointer transmutations, for example between `*const DialogFileBrowserOptions` and `*const sys::DialogsFileBrowserOptions` which may come in handy in some cases.

crates/sys/src/lib.rs
```diff
@@ -29,12 +29,18 @@ mod inlines;
...
+
+/// Create a static C string of type [`*const c_char`][core::ffi::c_char]
+/// referring to a `'static` string.
+/// Will automatically add a NUL terminator.
+// TODO: don't produce intermediate `CStr` whose `length` part we don't use
 #[macro_export]
 macro_rules! c_string {
     ($str:expr $(,)?) => {{
-        concat!($str, "\0").as_ptr() as *const core::ffi::c_char
+        $crate::cstr!($str).as_ptr()
```
str4d, line 43: Something about this change broke the logging macros. From CI (compiling the examples):
```
error: unexpected token
  --> flipperzero/examples/hello-rust.rs:31:5
   |
31 |     info!("Hello, reader of the logs!");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `$crate::log` which comes from the expansion of the macro `info` (in Nightly builds, run with -Z macro-backtrace for more info)

error: unexpected token
  --> flipperzero/examples/hello-rust.rs:35:5
   |
35 |     debug!("Return code: {}", ret_code);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `$crate::log` which comes from the expansion of the macro `debug` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `flipperzero` (example "hello-rust") due to 2 previous errors
```

JarvisCraft: I took me some time to find the issue:

It seems that the (imported) `cstr` macro requires the value to be literal (this is logical) which is not true when using `module_path!`, unlike our old `c_string` which (I guess, magically) correctly expands `concat!(module_path!(..), "\0")`.

My suggestion here is to keep the changes, since they are less error-prone due to strictire compile-time validations but provide our convenient macros for things like `module_path` producing valid ptrs, although I am mnot as sure if it is the right approach.

JarvisCraft: I've pushed my vision of this change which reworks `log` macro, although there may be other ways to do it

str4d: I don't see any relevant changes pushed to this PR.

JarvisCraft: My bad, just pushed the change in 13cca7ae9ead9e1212c270d3c0ae2c8af7b34d42

str4d:
Build is failing.

JarvisCraft:
> Build is failing.

Yup, I am going to fix that, but now cstr-literals are [stabilized][cstr-literal-stabilization] in 1.76 I may rework this PR to omit the redundant dependency.

[cstr-literal-stabilization]: https://github.com/rust-lang/rust/pull/117472

JarvisCraft:
Since this PR effectively depends on #120, the Clippy should be fixed by the latter.

PR #93, opened 2023-07-24
Update to SDK 34.3 (Firmware 0.87.0)

avisbisk:
Followed the docs to generate bindings and get this working locally on my system. Not sure if this is helpful, but figured I would create a PR incase this is. I didn't update the crate version in README, but I added the process changes that I needed to the updating-sdk docs. 
There was also a change with mod.rs, not sure what the root of the change was but looks like dolphin_deed only accepts one param as the deed.

### DIFF
--- a/crates/flipperzero/src/dolphin/mod.rs
+++ b/crates/flipperzero/src/dolphin/mod.rs
@@ -28,7 +28,7 @@ impl Dolphin {
     ///
     /// In future it will become part of assets. Thread safe, async.
     pub fn deed(&mut self, deed: Deed) {
-        unsafe { sys::dolphin_deed(self.data.as_ptr(), deed.to_raw()) };
+        unsafe { sys::dolphin_deed(deed.to_raw())};
     }
 
     /// Retrieves the dolphin's current stats.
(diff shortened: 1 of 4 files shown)


crates/flipperzero/src/dolphin/mod.rs
```diff
@@ -28,7 +28,7 @@ impl Dolphin {
     ///
     /// In future it will become part of assets. Thread safe, async.
     pub fn deed(&mut self, deed: Deed) {
-        unsafe { sys::dolphin_deed(self.data.as_ptr(), deed.to_raw()) };
+        unsafe { sys::dolphin_deed(deed.to_raw())};
```
JarvisCraft, line 31:
```suggestion
        unsafe { sys::dolphin_deed(deed.to_raw()) };
```

JarvisCraft (changes requested):

str4d:
I'll review this tomorrow (now that I have some time again for Flipper Zero stuff).

JarvisCraft:
@str4d, I will soon open the actualized version of this PR with for `0.88`

PR #95, opened 2023-08-11
Update test harness for upstream changes to loader

str4d:
This now enables test filtering:
```
$ cd crates/
$ cargo test -- FILTER ANOTHER_FILTER
```

### DIFF
--- a/crates/test/macros/src/lib.rs
+++ b/crates/test/macros/src/lib.rs
@@ -124,15 +124,17 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
                 ret
             }
 
-            fn test_list() -> impl Iterator<Item = (&'static str, &'static str, ::flipperzero_test::TestFn)> {
+            fn test_list() -> impl Iterator<Item = (&'static str, &'static str, ::flipperzero_test::TestFn)> + Clone {
                 let ret = ::core::iter::empty();
                 #( let ret = ret.chain(#test_lists); )*
                 ret
             }
 
             // Test runner entry point
-            fn main(_args: *mut u8) -> i32 {
-                match ::flipperzero_test::__macro_support::run_tests(test_count(), test_list()) {
+            fn main(args: *mut u8) -> i32 {
+                // SAFETY: Flipper Zero passes arguments to FAPs as a C string.
+                let args = unsafe { ::flipperzero_test::__macro_support::Args::parse(args) };
+                match ::flipperzero_test::__macro_support::run_tests(test_count(), test_list(), args) {
                     Ok(()) => 0,
                     Err(e) => e,
                 }
--- a/crates/test/src/lib.rs
+++ b/crates/test/src/lib.rs
@@ -65,7 +65,7 @@ impl ufmt::uDisplay for TestFailure {
 }
 
 pub mod __macro_support {
-    use core::ffi::c_char;
+    use core::ffi::{c_char, CStr};
 
     use flipperzero_sys as sys;
     use sys::furi::UnsafeRecord;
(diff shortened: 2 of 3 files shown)


str4d:
The syntax `cargo test FILTER` doesn't work, because that requires every crate in the workspace to have a `main()` when the test harness is disabled (and we need it disabled for the `flipperzero` tests to work). You can use either `cargo test -p flipperzero FILTER` or `cargo test -- FILTER` instead.

crates/test/macros/src/lib.rs
```diff
@@ -124,15 +124,16 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
                 ret
             }
 
-            fn test_list() -> impl Iterator<Item = (&'static str, &'static str, ::flipperzero_test::TestFn)> {
+            fn test_list() -> impl Iterator<Item = (&'static str, &'static str, ::flipperzero_test::TestFn)> + Clone {
```
JarvisCraft, line 127: Since we are in a macro context here, I think that we should use fully qualified path to `Iterator` and `Clone`

crates/test/src/lib.rs
```diff
@@ -74,6 +74,49 @@ pub mod __macro_support {
 
     const RECORD_STORAGE: *const c_char = sys::c_string!("storage");
 
+    pub fn parse_args<'a>(args: *mut u8) -> Args<'a> {
+        if args.is_null() {
+            Args("")
+        } else {
+            let args_cstr = unsafe { CStr::from_ptr(args.cast_const().cast()) };
```
JarvisCraft, line 81: ATM the function is unsound since an invalid pointer can be passed to it (it is not `unsafe`) and yet the `args` has a number of [restrictions](https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#safety) on it.

It should either be validated or the function should be marked `unsafe`.

JarvisCraft (changes requested):

crates/test/macros/src/lib.rs
str4d: I don't believe that to be necessary, as both of these are part of [`core::prelude`](https://doc.rust-lang.org/core/prelude/v1/index.html), and are therefore available directly in every `no-std` Rust program.

crates/test/src/lib.rs
str4d: IIRC my thinking was that this is validated because we're paying in a pointer from the Flipper Zero that we know that (or rather, by definition) the pointer is to a valid C string. However, it's true that because I moved this into a macro helper, it could be used by others. So yes, I'll make this `unsafe`.

crates/test/macros/src/lib.rs
JarvisCraft: AFAIK, there are at least two ways to break this i.e.:
* `#![no_implicit_prelude]`
* having similarly named traits in the scope.

While the former is quite exotic, the latter may appear thus I would prefer having a bit more hygienic fix for this.

Arguing against myself, it still is not 100%-hygienic, but to me it looks like a more error-prone solution.

crates/test/macros/src/lib.rs
```diff
@@ -124,15 +124,17 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
...
                 let ret = ::core::iter::empty();
                 #( let ret = ret.chain(#test_lists); )*
                 ret
             }
 
             // Test runner entry point
-            fn main(_args: *mut u8) -> i32 {
-                match ::flipperzero_test::__macro_support::run_tests(test_count(), test_list()) {
+            fn main(args: *mut u8) -> i32 {
```
JarvisCraft, line 134: Being a bit of a devils advocate here: `main` also, seems to be `unsafe` since the caller is again required to provide a valid pointer.

The same logic works here as with `parse`: since it is a generated function, it is accessible to nearby code which may, for example, call it with null.

Not to be too copy-pasty here, something like

```markdown
# Safety

This should only be called by Flipperzero OS
and never manually.
```

would be enough.

JarvisCraft (changes requested):

crates/test/macros/src/lib.rs
str4d: Let's address this in a separate PR then, that goes through all the macros and ensure they follow a consistent pattern.

JarvisCraft (approved):

crates/test/macros/src/lib.rs
str4d: Opened #98.

Issue #89, opened 2023-06-21
Add support for Fast FAP loading


str4d:
https://github.com/flipperdevices/flipperzero-firmware/pull/2790 bumps the SDK to 31 with 15x faster loading of FAP files. It requires adding `.fast.rel` sections, so we will need to alter `flipperzero-rt` to enable it.

dcoles:
Looking at the PR, I think what it's doing is creating a hash table for symbol lookup, rather than a linear search.

This might need to be done as some sort of post-processing step as I don't think it's possible to do this in `build.rs` or the [linker script](https://github.com/flipperzero-rs/flipperzero/blob/main/crates/rt/flipperzero-rt.ld). This appears to be what [`scripts/fastfap.py`](https://github.com/flipperdevices/flipperzero-firmware/pull/2790/files#diff-a0e50a592b3e018edf7898a32074ccd786babb9553d3099ac7920c24b8ef983e) is for.

DrZlo13:
Yep, [fastfap.py](https://github.com/flipperdevices/flipperzero-firmware/pull/2790/files#diff-a0e50a592b3e018edf7898a32074ccd786babb9553d3099ac7920c24b8ef983e) is post-processing elf files not only for hashing, but also to create an optimized relocation structure for our elf loader. This removes a large number of calls to the storage, and that's how such an acceleration is achieved.

I recommend you to wait until this feature is released because the format of these sections is subject to change. I also left a fallback to the old usual linking process if optimized sections is not present.

dcoles:
@DrZlo13 Thanks for the information. Definitely appreciate the fallback, so this isn't an urgent thing to address.

str4d:
This was released in firmware version 0.86.1, and once #93 merges we will require at least 0.87.0, so we can start working on this.

str4d:
This is now implemented in #99. I _think_ I got the encodings right, and it appears to work (`cargo test` gets noticeably faster) but I'm unsure if the MD5 for tempfile names in `fastfap.py` is just for convenience or meant to be load-bearing.

DrZlo13:
@str4d MD5 for temporary filenames is only for reducing filename size. Not all filesystems can handle files named from a C++ generated section name.

str4d:
@DrZlo13 thanks, that clarifies things (I didn't know `objcopy` derives section names from filenames in this way).

[...]

PR #99, opened 2023-08-25
tools: Add `fap-lld` wrapping linker that generates `.fast.rel` sections

str4d:
Closes flipperzero-rs/flipperzero#89.


str4d:
In local testing on my Macbook Air M1, this reduces `cargo test` runtimes from 13s to 11.7s.

str4d:
From the error in the example build logs, it looks like we can't use whatever `objcopy` is in `PATH`, and will probably need to just use `llvm-objcopy` and tell people to `rustup component add llvm-tools` to get FastFAP support.

crates/.cargo/config.toml
```diff
@@ -1,4 +1,5 @@
 [target.thumbv7em-none-eabihf]
+linker = "./fap-lld.py"
```
str4d, line 2: Unlike the `runner` config, I could not persuade `cargo` to accept a multi-argument string here. So I'm relying on `/usr/bin/env` existing.

If we want the speed optimisations to be available to users, we probably need to consider cleaning up and publishing the tools, so downstream users can install them and set this to e.g. `flipperzero-lld`.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/6e2f492271acf85881fc8b1fa55f81a0fbfb0da1..0f7e39751523840e65adf031f33af9181db46c06) to fix merge conflicts caused by #97 being merged (which I had been holding off merging, because dependabot will perform these kinds of rebases automatically on its own PRs).

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/0f7e39751523840e65adf031f33af9181db46c06..8f964b2410c71b912a309a5baa961aab39ec1075) to use deterministic filenames for the `.fast.rel` section data.

crates/.cargo/config.toml
dcoles: I've been thinking about how we handle required binaries like `fap-lld`. It's probably not unreasonable to ask people to `cargo install flipperzero-tools` (or `flipperzero-lld`) if they want to develop applications.

I'm not if there's a way to specify a `cargo install` dependency, but it seems like something you could check in `build.rs`.

crates/fap-lld.py
```diff
@@ -0,0 +1,33 @@
...
+            '--quiet',
+            '--release',
+            '--bin',
+            'fap-lld',
+            '--',
+        ] + args,
+        cwd=os.path.join(os.path.dirname(__file__), TOOLS_PATH),
+    )
+    if result.returncode:
+        sys.exit(result.returncode)
```
dcoles, line 29: For a completed process, there should always be a valid `returncode`, so it's OK to call this unconditionally.

tools/src/bin/fap-lld/fastrel.rs
```diff
@@ -0,0 +1,156 @@
...
+    name: &'data str,
+}
+
+impl<'data> FastRel<'data> {
+    fn gnu_sym_hash(&self) -> u32 {
+        let mut h = 0x1505;
+        for c in self.name.as_bytes() {
+            h = (h << 5) + h + u32::from(*c);
+        }
+        h
```
dcoles, line 87: This is a nit-picky personal style thing: I find it helpful to have a blank line before the return value.

Visually it makes it a bit more distinct, which I find helpful when scanning a function. Thoughts?

tools/src/bin/fap-lld/fastrel.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        Ok(FastRelSection {
+            name: format!(".fast{}", section_name),
+            fastrel_offsets,
+        })
+    }
+
+    fn write(&self, mut w: impl Write) -> io::Result<()> {
+        const VERSION: u8 = 1;
```
dcoles, line 134: It might be useful to pull this out to the module level, just so it's very clear what version this module supports.

tools/src/bin/fap-lld/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+fn main() -> Result<(), Error> {
+    // Run the real linker with the given arguments.
+    let res = Command::new("rust-lld")
+        .args(env::args_os().skip(1))
+        .status()?;
+    if !res.success() {
+        process::exit(res.code().unwrap_or(-1));
```
dcoles, line 39: Similar to the Python code, [`ExitStatus::code`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code) will give `Some(0)` on success, so we can remote the [`ExitStatus::success`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.success) conditional.

dcoles (approved):
This looks great!

None of my comments are blockers, so feel free to merge when you're ready.

crates/.cargo/config.toml
str4d: `cargo` has [unstable support for artifact dependencies](https://doc.rust-lang.org/cargo/reference/unstable.html#artifact-dependencies), which could be a way to use this in future. But most embedded devs are used to needing a separate tooling package, so I think that's fine to end up requiring. For our own development we can continue to use the Python wrappers for convenience (and to ensure we are using the development version of the tooling instead of any installed version).

tools/src/bin/fap-lld/fastrel.rs
str4d: I usually do that as well, but this is such a small function that it didn't seem worthwhile. If I were altering this, I'd probably rewrite it to use `Iterator::fold` instead.

tools/src/bin/fap-lld/main.rs
str4d: Here we cannot, because if this succeeds, we want to continue on and perform the post-linker optimizations.

PR #101, opened 2023-08-28
Update SDK API to 35.0 (firmware 0.89.0)

dcoles:
Updated bindings using the following command:

```
docker run --rm $(docker build --build-arg BRANCH=0.89.0 -q -f tools/Dockerfile .) > crates/sys/src/bindings.rs
````

There is a 0.90.0rc just around the corner, but it has a compatible SDK (35.1).

### DIFF
--- a/crates/sys/src/bindings.rs
+++ b/crates/sys/src/bindings.rs
@@ -110,7 +110,7 @@ impl<T> ::core::fmt::Debug for __IncompleteArrayField<T> {
         fmt.write_str("__IncompleteArrayField")
     }
 }
-pub const API_VERSION: u32 = 2228227;
+pub const API_VERSION: u32 = 2293760;
 pub type wint_t = core::ffi::c_int;
 pub type _off_t = core::ffi::c_long;
 pub type _fpos_t = core::ffi::c_long;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/sys/src/bindings.rs [hunk 1]
```diff
@@ -5068,6 +5086,56 @@ extern "C" {
     #[doc = "Decrypt data\n\nReturns:\n\n* true on success\n\n# Arguments\n\n* `input` - pointer to input data\n* `output` - pointer to output data\n* `size` - input/output buffer size in bytes\n\n"]
     pub fn furi_hal_crypto_decrypt(input: *const u8, output: *mut u8, size: usize) -> bool;
 }
+extern "C" {
+    #[doc = "Encrypt the input using AES-CTR\nDecryption can be performed by supplying the ciphertext as input. Inits and deinits the AES engine internally.\n\nReturns:\n\n* true on success\n\n# Arguments\n\n* `key` - [Direction: In] pointer to 32 bytes key data\n* `iv` - [Direction: In] pointer to 12 bytes Initialization Vector data\n* `input` - [Direction: In] pointer to input data\n* `output` - [Direction: In, Out] pointer to output data\n* `length` - length of the input and output in bytes\n\n"]
+    pub fn furi_hal_crypto_ctr(
```
str4d, line 5091: I'll add toolbox wrappers for these new AES functions in a follow-up PR.

str4d (approved):
utACK. I did not generate the update myself to confirm, but the changes look as I anticipated they would.

PR #104, opened 2023-09-02
tools: Migrate to Protobuf RPC sessions

### LINKED ISSUE #76: tools: Migrate from text-based CLI to Protobuf RPC
Currently the binaries in `tools/` use the text-based CLI over the serial port to communicate with a Flipper Zero and run FAPs. We should instead use the `start_rpc_session` CLI command to drop into the [Protobuf RPC control mode](https://github.com/flipperdevices/flipperzero-protobuf), using e.g. the `prost` crate. This would remove most of the hand-written CLI parsing logic, and probably be more accurate and flexible.

In particular, I am curious about whether [`PB_App.StartRequest`](https://github.com/flipperdevices/flipperzero-protobuf/blob/f71c4b7f750f2539a1fed08925d8da3abdc80ff9/application.proto#L6C18-L9) is specifically a request to `Applications` that enables us to provide arguments to FAPs (which the text CLI `loader open Applications foo.fap` does not), or whether it is a request to `loader` (and thus the same as the text CLI).

str4d:
Closes #76.


tools/src/storage.rs
```diff
@@ -145,124 +122,117 @@ impl FlipperStorage {
...
     }
 
-    /// Format external storage.
-    pub fn format_ext(&mut self) -> io::Result<()> {
-        self.cli.send_and_wait_eol("storage format /ext")?;
-        self.cli.send_and_wait_eol("y")?;
-        self.cli.consume_response()?;
+    // /// Format external storage.
+    // pub fn format_ext(&mut self) -> io::Result<()> {
+    //     self.cli.send_and_wait_eol("storage format /ext")?;
```
str4d, line 202: AFAICT there is no RPC equivalent to `storage format`.

tools/src/bin/run-fap.rs
```diff
@@ -62,17 +65,10 @@ impl From<io::Error> for Error {
...
-        if serial_cli
-            .consume_response()?
-            .contains("No application is running")
-        {
-            break Ok(());
-        }
-        thread::sleep(Duration::from_millis(200));
-    }
+fn wait_for_idle(session: &mut RpcSession) -> io::Result<()> {
+    // TODO: need equivalent of "loader info" CLI command.
```
str4d, line 69: There is probably an equivalent to this (to determine whether an app is still running), but I have not figured out what it is yet.

str4d:
`--bin storage -- list` seems to be slower than the text CLI version. Not sure there's much we can do about that.

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/b9905531d2e7a624d2e67d7d0a3526c78d248e03..87cbc4d2dccf7e0511b73520925e84b2e1c1157f).

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/87cbc4d2dccf7e0511b73520925e84b2e1c1157f..22747a941fa83dda10d6af836921460f6698a208) to clean up the protobuf generation and move the output files into a submodule, making room for other submodules under `proto` for organising the async impl.

PR #106, opened 2023-09-07
Upgrade to sdk 38.0

Nereuxofficial:
It would probably good to make a dev branch for this as it has not been extensively tested. 

I used this awesome crate to play a bit with the flipper zero and it's really cool that this crate enables that without external bindings! Though the SDK Version was incompatible so i updated it. I also updated the documentation for some examples as I went through them

### DIFF
--- a/crates/flipperzero/examples/threads.rs
+++ b/crates/flipperzero/examples/threads.rs
@@ -1,4 +1,5 @@
-//! Demonstrates use of threads on the Flipper Zero.
+//! Demonstrates use of threads on the Flipper Zero by spawning two threads and dropping the first +
+//! while waiting for the second to complete.
 
 #![no_main]
 #![no_std]
--- a/crates/sys/src/bindings.rs
+++ b/crates/sys/src/bindings.rs
@@ -110,7 +110,7 @@ impl<T> ::core::fmt::Debug for __IncompleteArrayField<T> {
         fmt.write_str("__IncompleteArrayField")
     }
 }
-pub const API_VERSION: u32 = 2293760;
+pub const API_VERSION: u32 = 2490368;
 pub type wint_t = core::ffi::c_int;
 pub type _off_t = core::ffi::c_long;
 pub type _fpos_t = core::ffi::c_long;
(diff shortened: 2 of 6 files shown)


Nereuxofficial:
Also a small observation: I saw cargo-runner.py use `run-fap` from the tools. Could this be used as a runner for the flipperzero-template to avoid uploading and executing manually, improving developer experience? If so, I'd love to implement that!

crates/sys/src/bindings.rs
```diff
@@ -6011,7 +6030,7 @@ extern "C" {
     pub fn furi_hal_region_is_provisioned() -> bool;
 }
 extern "C" {
-    #[doc = "Get region name\n2 letter Region code according to iso 3166 standard There are 2 extra values that we use in special cases:\n* \"00\" - developer edition, unlocked\n* \"WW\" - world wide, region provisioned by default\n* \"--\" - no provisioned region\n\nReturns:\n\n* Pointer to string\n\n"]
+    #[doc = "Get region name\n2 letter Region code according to iso 3166 standard There are 2 extra values that we use in special cases: RM, whats the reason you not doing a release? Waiting for my commits?\n* \"00\" - developer edition, unlocked\n* \"WW\" - world wide, region provisioned by default\n* \"--\" - no provisioned region\n\nReturns:\n\n* Pointer to string\n\n"]
```
dcoles, line 6033: Which branch/tag are you building these bindings against? There seems to be a number of comments/symbols exported that I can't find in the upstream [flipperdevices/flipperzero-firmware](https://github.com/flipperdevices/flipperzero-firmware) repository.

**Note:** We do not officially support non-standard firmware, since it would be a compatibility nightmare; but you can always checkout this repository, regenerate the bindings and then set a [path dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) in your own project. This should be a lot less painful once #41 is fixed and we have Cargo build the user's preferred bindings on demand.

crates/flipperzero/examples/threads.rs
```diff
@@ -1,4 +1,5 @@
-//! Demonstrates use of threads on the Flipper Zero.
+//! Demonstrates use of threads on the Flipper Zero by spawning two threads and dropping the first +
+//! while waiting for the second to complete.
```
dcoles, line 2: Thank you for making improvements to the documentation for these examples.

dcoles (changes requested):

dcoles:
> Also a small observation: I saw cargo-runner.py use `run-fap` from the tools. Could this be used as a runner for the flipperzero-template to avoid uploading and executing manually, improving developer experience? If so, I'd love to implement that!

Absolutely! That's a great idea!

We've considering making the tools a requirement for most development. The `cargo-runner.py` script was an attempt to have things work automatically; but it never quite worked 100% (especially if you wanted to support multiple OSs). So my recommendation would be to just set [the runner](https://github.com/flipperzero-rs/flipperzero/blob/4406df56071902765b78163ddfa561a3660c5734/crates/.cargo/config.toml#L3) to `run-fap` directly.

crates/sys/src/bindings.rs
Nereuxofficial: Sorry it was the unleashed firmware. I had assumed they would be API-compatible. I now used the main branch and made another branch for that. Thanks for the heads up!

dcoles (approved):
I updated the bindings to 38.0 since the [current release candidate for 0.91](https://github.com/flipperdevices/flipperzero-firmware/blob/0.91.3-rc/firmware/targets/f7/api_symbols.csv#L2) is now up to that value.

dcoles:
Thanks for the PR!

PR #107, opened 2023-09-08
Github action for "nightly" builds of the bindings

Nereuxofficial:
I made a short workflow that checks out the newest Flipper Zero Mainline firmware, builds bindings against it and pushes it into a "nightly" branch.

This is more of a proposal. Using this would allow:
- Users to use bindings built against the git version of the flipper zeros firmware via:
```toml
[dependencies]
flipperzero = { git="https://github.com/flipperzero-rs/flipperzero", package="flipperzero", branch="nightly"}
flipperzero-sys = { git="https://github.com/flipperzero-rs/flipperzero", package="sys", branch="nightly" }
flipperzero-rt = { git="https://github.com/flipperzero-rs/flipperzero", package="sys", branch="nightly" }
flipperzero-alloc = { git="https://github.com/flipperzero-rs/flipperzero", package="sys", branch="nightly" }
```
And they could also specify a specific commit to pin the dependency.
- Early discovery of API-changes to bindings
- Fully automatic builds against the newest SDK potentially also useable for custom firmware or at least easier to adjust to custom firmware

### DIFF
--- a/crates/flipperzero/Cargo.toml
+++ b/crates/flipperzero/Cargo.toml
@@ -35,6 +35,10 @@ lock_api = "0.4"
 digest = "0.10"
 bitflags = "1.0"
 
+# Embedded-hal
+embedded-hal = { version = "1.0.0-rc.1", optional = true }
+embedded-hal-0 = { package = "embedded-hal", version = "0.2.7", features = ["unproven"], optional = true }
+
 # Docs
 document-features = { workspace = true, optional = true }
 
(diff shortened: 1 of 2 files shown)


dcoles:
At present we're using the [`Dockerfile` in tools](https://github.com/flipperzero-rs/flipperzero/blob/main/tools/Dockerfile) to generate bindings in a relatively reproducible way.

I need to think a little more about the implications of creating a "nightly" branch like this (it's really [flipper-firmware@dev](https://github.com/flipperdevices/flipperzero-firmware/) nightly), especially if folks are going to start depending upon it.

Perhaps a better approach would be to generate a binary artefact that can be easily downloaded, a bit like what is done on [updates.flipperzero.one](https://update.flipperzero.one/builds/firmware/dev/). Though I do feel that moving away from checking-in the bindings is going to be the better option long-term, so folks aren't pinned to a specific SDK.

PR #111, opened 2023-10-06
flipperzero: I2C interface

str4d:
This adds a Rust API for using both the external and internal I2C bus with 7-bit addresses.

### DIFF
--- a/crates/flipperzero/src/gpio/i2c.rs
+++ b/crates/flipperzero/src/gpio/i2c.rs
@@ -1,0 +1,342 @@
+//! I2C interface for the Flipper Zero.
+
+use core::fmt;
+
+use flipperzero_sys as sys;
+
+use crate::furi::time::Duration;
+
+/// The address of the internal LED controller.
+///
+/// This is an [LP5562].
+///
+/// [LP5562]: https://www.ti.com/lit/ds/symlink/lp5562.pdf
+pub const INTERNAL_LED_CONTROLLER: DeviceAddress = DeviceAddress::new(0x30);
+
+/// The address of the internal battery fuel gauge.
+///
+/// This is a [BQ27220].
+///
+/// [BQ27220]: https://www.ti.com/lit/ds/symlink/bq27220.pdf
+pub const INTERNAL_BATTERY_FUEL_GAUGE: DeviceAddress = DeviceAddress::new(0x55);
+
+/// The address of the internal battery charger.
+///
+/// This is a [BQ25896].
+///
+/// [BQ25896]: https://www.ti.com/lit/ds/symlink/bq25896.pdf
+pub const INTERNAL_BATTERY_CHARGER: DeviceAddress = DeviceAddress::new(0x6B);
+
+/// A 7-bit I2C device address.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct DeviceAddress(u8);
+
+impl fmt::Debug for DeviceAddress {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("DeviceAddress")
+            .field(&(self.0 >> 1))
+            .finish()
+    }
+}
+
+impl ufmt::uDebug for DeviceAddress {
+    #[inline]
+    fn fmt<W>(&self, f: &mut ufmt::Formatter<'_, W>) -> Result<(), W::Error>
+    where
+        W: ufmt::uWrite + ?Sized,
+    {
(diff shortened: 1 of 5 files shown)


str4d:
For now this just uses the helper interfaces that the Flipper Zero SDK provides for 7-bit addresses, but those are all themselves implemented in terms of the underlying 10-bit-capable raw interface which is also exposed. So if that's desired then we could re-implement the helper methods in Rust. However, `furi_hal_i2c_is_device_ready` is only exposed for 7-bit addresses.

crates/flipperzero/src/gpio/i2c.rs
```diff
@@ -0,0 +1,402 @@
...
+    ) -> Result<(), Error> {
+        if unsafe { sys::furi_hal_i2c_write_reg_16(self.handle, device.0, reg_addr, data, timeout) }
+        {
+            Ok(())
+        } else {
+            Err(Error::TransferFailed)
+        }
+    }
+
+    /// Writes the given data into `device`'s memory starting at `mem_addr`.
```
dcoles, line 348: I might be missing something, but it looks like `Bus` provides almost exactly the same methods as `BusHandle`.

It makes it a bit tricky to tell what is the correct way to use the API. I would normally expect an API that looked something like the following:

```rust

const i2c::Bus::INTERNAL: &mut sys::FuriHalI2cBusHandle = &mut sys::furi_hal_i2c_handle_power;

{
    let bus = i2c::Bus::acquire(i2c::Bus::INTERNAL);

    if bus.is_device_ready(i2c::DeviceAddress::new(0x68), 0) {
        ...
    }

    // Bus is released when dropped
}
```

crates/flipperzero/src/gpio/i2c.rs
```diff
@@ -0,0 +1,402 @@
...
+    ///
+    /// `timeout` is in milliseconds.
+    pub fn write_all(
+        &mut self,
+        device: DeviceAddress,
+        mem_addr: u8,
+        data: &mut [u8],
+        timeout: u32,
```
dcoles, line 356: It would be nice to use `Duration` for the timeouts, rather than bare integers.

dcoles (commented):
Very cool! Looks like I'll need to find me a DS3231 for testing.

crates/flipperzero/src/gpio/i2c.rs
str4d: That's what I originally did, but then I looked into how I2C was used elsewhere by Flipper Zero, and it seemed like only acquiring as long as you need it was the way to go sometimes? But then I figured some issue would want to acquire for longer periods. So the `Bus` methods are just simple wrappers when you only need to call something once.

All this is to say, I'd love some input from someone who works with I2C more than I have (never).

crates/flipperzero/src/gpio/i2c.rs
str4d: Slightly annoyingly, the most recent SDK update fixed the documentation for these methods to indicate that these are actually in milliseconds instead of ticks. So using `Duration` would add two conversions (one outside the method, and one inside) on top of the presumed third conversion happening inside the SDK.

str4d: Inside the SDK it results in a div for an assertion, and a couple of mults. That's also what we end up adding via `Duration`, plus some FFI calls to get the tick frequency twice. Overall it's probably not too costly, so I'll do that and we can improve this via SDK changes later.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/0dfbaa4bcde82b348a17b354469731fe8b8b2cd3..36801aff3bfe7f10884214475f9b7801b833fadd) to use `Duration` instead of `u32`.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/36801aff3bfe7f10884214475f9b7801b833fadd..c42d5ca53ba9bfa1e425040cdeddb97389978dad) to remove the convenience methods from `i2c::Bus`.

crates/flipperzero/src/gpio/i2c.rs
str4d: I've removed the convenience methods from `i2c::Bus` to simplify this PR. Users can do the same thing directly by calling `i2c::Bus::with_handle`.

dcoles (approved):
LGTM! 👍

str4d:
At some point in the future I'll experiment with 10-bit interface support, either once I have a 10-bit device to test with, or as part of adding `embedded-hal` support for #113. But this is fine to merge for now.

PR #117, opened 2023-10-26
Add embedded-hal I2C traits implementations

agarof:
Implements the I2C traits from embedded-hal, both the 0.2 and 1.0 rc1 versions.

Both implementations are optional features to avoid forcing the dependencies onto users.

Tested quickly using a logic analyzer and the [ssd1306 crate](https://crates.io/crates/ssd1306) with the following code:
```rust
use core::fmt::Write;
use ssd1306::prelude::*;

let bus = flipperzero::gpio::i2c::EmbeddedHalBus::new(
    flipperzero::gpio::i2c::Bus::EXTERNAL,
    flipperzero::furi::time::Duration::from_millis(100),
);

let interface = ssd1306::I2CDisplayInterface::new(bus);
let mut display = ssd1306::Ssd1306::new(interface, DisplaySize128x32, DisplayRotation::Rotate0)
    .into_terminal_mode();

display.init().unwrap();
display.clear().unwrap();

display.write_str("Hello from the\nFlipper Zero\n").unwrap();
```
image

### DIFF
--- a/crates/flipperzero/src/gpio/i2c.rs
+++ b/crates/flipperzero/src/gpio/i2c.rs
@@ -134,8 +134,6 @@ impl BusHandle {
     }
 
     /// Enumerates the devices that are present and ready on this bus.
-    ///
-    /// `per_device_timeout` is in milliseconds.
     pub fn enumerate_devices(
         &mut self,
         per_device_timeout: Duration,
@@ -149,8 +147,6 @@ impl BusHandle {
 
     /// Checks if the device with address `i2c_addr` is present and ready on the bus.
     ///
-    /// `timeout` is in milliseconds.
-    ///
     /// Returns `true` if the device is present and ready, false otherwise.
     pub fn is_device_ready(&mut self, device: DeviceAddress, timeout: Duration) -> bool {
         unsafe {
@@ -159,8 +155,6 @@ impl BusHandle {
     }
 
     /// Reads the 8-bit register at `reg_addr` on `device`.
-    ///
-    /// `timeout` is in milliseconds.
     pub fn read_u8(
         &mut self,
         device: DeviceAddress,
@@ -184,8 +178,6 @@ impl BusHandle {
     }
 
     /// Reads the 16-bit register at `reg_addr` on `device`.
-    ///
-    /// `timeout` is in milliseconds.
     pub fn read_u16(
         &mut self,
         device: DeviceAddress,
@@ -209,8 +201,6 @@ impl BusHandle {
     }
 
     /// Reads `device`'s memory starting at `mem_addr` into the given buffer.
-    ///
-    /// `timeout` is in milliseconds.
     pub fn read_exact(
         &mut self,
         device: DeviceAddress,
@@ -235,8 +225,6 @@ impl BusHandle {
     }
 
     /// Writes the given value into the 8-bit register at `reg_addr` on `device`.
-    ///
-    /// `timeout` is in milliseconds.
     pub fn write_u8(
         &mut self,
         device: DeviceAddress,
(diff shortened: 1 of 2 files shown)


crates/flipperzero/src/gpio/i2c.rs
```diff
@@ -309,13 +309,410 @@ impl BusHandle {
             Err(Error::TransferFailed)
         }
     }
+
+    /// Reads data from `device` and writes it to the `data` buffer.
+    ///
+    /// `timeout` is in milliseconds.
```
dcoles, line 315: No longer relevant because you use `Duration`.

crates/flipperzero/src/gpio/i2c.rs
```diff
@@ -309,13 +309,410 @@ impl BusHandle {
...
+        unsafe {
+            sys::furi_hal_i2c_tx(
+                self.handle,
+                device.0,
+                data.as_ptr(),
+                data.len(),
+                timeout.as_millis() as u32,
+            )
+        }
+        .then_some(())
```
dcoles, line 331: I didn't know about this method on `bool`. Cool!

crates/flipperzero/src/gpio/i2c.rs
```diff
@@ -309,13 +309,410 @@ impl BusHandle {
...
+    /// Read data into the provided buffer
+    Read(&'a mut [u8]),
+    /// Write data from the provided buffer
+    Write(&'a [u8]),
+}
+
+// These exist to allow compatibility with both embedded_hal 1.0 and 0.2 versions of the Operation
+// enum
+
+enum OperationKind {
```
dcoles, line 525: Does it makes sense for these enums to derive the standard `Debug`, `Copy`, `Clone`, `Eq`, `PartialEq`, `Hashable` traits?

dcoles (approved):
Hi @agarof. Thanks for the pull request! It looks great and would be happy to accept it. I've wanted to implement parts of `embedded-hal` for a while, but just haven't had the time. Opens up a lot more libraries in the ecosystem.

Only feedback on the review is some stale comments and some optional derives that could be implemented.

crates/flipperzero/src/gpio/i2c.rs
agarof: Because `OperationKind` is only meant for internal use, I don't think it needs them.
I went ahead and derived `Debug`, `Eq` and `PartialEq` for `Operation`, mirroring the version found in [embedded-hal](https://docs.rs/embedded-hal/1.0.0-rc.1/src/embedded_hal/i2c.rs.html#279).

dcoles: Sounds reasonable to me.

crates/flipperzero/src/gpio/i2c.rs
dcoles: Yes please. That would be appreciated (I didn't see that it was also on the other `Duration` methods).

agarof: Done

dcoles:
@agarof Thanks for the pull request. Let me know if you're planning any other related PRs that should be in the next release, so I hold off on making it until then.

agarof:
@dcoles I should have time to implement `embedded-hal`'s [digital](https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/digital/index.html) traits for the GPIOs during the week.

str4d:
I'm glad you're going to work on this, because I'd just started a few weeks ago integrating `embedded-hal` myself (at part of writing SPI bindings), and was running into issues with the digital IO traits.

PR #122, opened 2023-12-09
flipperzero: Scene management

str4d:


str4d:
@JarvisCraft I put this together as an initial context-less (for simplicity) approach (as part of trying to get the NFC bindings working), then saw that the view dispatcher (my actual goal for testing the NFC APIs) already has bindings inside your GUI PR #29. Do you think the view dispatcher bindings would be reasonable to try and extract into their own PR, or are they tightly bound to the rest of the GUI bindings?

JarvisCraft:
> @JarvisCraft I put this together as an initial context-less (for simplicity) approach (as part of trying to get the NFC bindings working), then saw that the view dispatcher (my actual goal for testing the NFC APIs) already has bindings inside your GUI PR #29. Do you think the view dispatcher bindings would be reasonable to try and extract into their own PR, or are they tightly bound to the rest of the GUI bindings?

@str4d, as I remember, the `ViewDispatcher` bindings were the hardest ones to implement (and I still have some uncertanty about its soundness in the way I initially though about it) so divvorcing it into a separate WIP-PR while merging the GUI MVP seems reasonable. The difference from your suggestion here is that I would prefer most things *except* `ViewDispatcher`-ones to be merged (they already provide a way to create a GUI) and postpone `ViewDispatcher` as a more complicated one.

I still expect myself to come out with breaking changes around the already implemented GUI API as I continue messing with `ViewDispatcher` support, but this is less likely to be critical and follows our current release versioning.

Issue #100, opened 2023-08-27
Allow running MIRI


JarvisCraft:
# Introduction

While some Flipper abstractions are easy to port to Rust, some have complicated lifecycles and are tricky to be wrapped in safe abstractions. Current work done in #29 is an illustration of this point.

---

The most common problems appearing are

## Aliasing

Arguably, the most complicated and still volatile part of Unsafe Rust especially painful when ahving to work with APIs which originally provide shared access (first of all, callbacks).

This is also especially tricky artound `Box`-based allocations for which there are uniquness guarantees at *some point*.

## Concurrency & mmulti-threading

When and where do we have to insert synchronizations for soundness?

This is specifically important for callback-based APIs.

## Extern type trickery

See rust-lang/rfcs#1861 for what we may need ideally.

We may want to create formally unsized wrapper-types of form

```rust
#[repr(transparent)]
struct Foo {
    // this should be an extern (thus unsized) type
    // making this wrapper unsized
    raw: sys::Foo,
}
```

which would only be usable via pointers (and their derivatives, i.e. references) safely transmutable to/from raw pointers.

---

In order to facilitate the creation of such abstractions, more tools could be used.

str4d:
We're going to have a very hard time of this, because [Miri doesn't yet support arbitrary FFI calls](https://github.com/rust-lang/miri/issues/11). I've locally made changes that enable `MIRI_NO_STD=1 cargo miri test --target thumbv7em-none-eabihf` to at least run into this issue:
```
error: unsupported operation: can't call foreign function `furi_record_open` on OS `none`
   --> /path/to/flipperzero/crates/sys/src/furi.rs:112:19
    |
112 |             data: crate::furi_record_open(name) as *mut T,
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't call foreign function `furi_record_open` on OS `none`
    |
    = help: this is likely not a bug in the program; it indicates that the program performed an operation that the interpreter does not support
```
but given the sheer quantity of FFI calls we have, I don't think we can reasonably expect to mock them all out for Miri.

str4d:
Aha, I missed the `-Zmiri-extern-so-file=<path to a shared object file>` experimental flag. So Miri does support at least simple FFI calls that only have integer (or void) args and return. That won't help for the majority of the FFI calls in the Flipper Zero SDK though, which operate on pointers.

PR #123, opened 2023-12-20
flipperzero: Enable Miri to start running the tests

str4d:
It doesn't *successfully* run the tests because Miri doesn't yet support FFI, but these changes enable the following command to actually start:

    MIRI_NO_STD=1 cargo miri test --target thumbv7em-none-eabihf

Part of flipperzero-rs/flipperzero#100.

### DIFF
--- a/crates/sys/src/lib.rs
+++ b/crates/sys/src/lib.rs
@@ -5,13 +5,16 @@
 
 // Features that identify thumbv7em-none-eabihf.
 // Until target_abi is stable, this also permits thumbv7em-none-eabi.
-#[cfg(not(all(
-    target_arch = "arm",
-    target_feature = "thumb2",
-    target_feature = "v7",
-    target_feature = "dsp",
-    target_os = "none",
-    //target_abi = "eabihf",
+#[cfg(not(any(
+    all(
+        target_arch = "arm",
+        target_feature = "thumb2",
+        target_feature = "v7",
+        target_feature = "dsp",
+        target_os = "none",
+        //target_abi = "eabihf",
+    ),
+    miri
 )))]
 core::compile_error!("This crate requires `--target thumbv7em-none-eabihf`");
 
--- a/crates/sys/src/lib.rs
+++ b/crates/sys/src/lib.rs
@@ -4,7 +4,6 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 
 // Features that identify thumbv7em-none-eabihf.
-// Until target_abi is stable, this also permits thumbv7em-none-eabi.
 #[cfg(not(any(
     all(
         target_arch = "arm",
@@ -12,7 +11,7 @@
         target_feature = "v7",
         target_feature = "dsp",
         target_os = "none",
-        //target_abi = "eabihf",
+        target_abi = "eabihf",
     ),
     miri
 )))]
--- a/crates/test/macros/src/lib.rs
+++ b/crates/test/macros/src/lib.rs
@@ -131,13 +131,26 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
             }
 
             // Test runner entry point
-            fn main(args: Option<&::core::ffi::CStr>) -> i32 {
+            pub(super) fn main(args: Option<&::core::ffi::CStr>) -> i32 {
                 let args = ::flipperzero_test::__macro_support::Args::parse(args);
                 match ::flipperzero_test::__macro_support::run_tests(test_count(), test_list(), args) {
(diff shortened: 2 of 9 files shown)


crates/sys/src/lib.rs
```diff
@@ -5,14 +5,14 @@
...
-#[cfg(not(all(
+#[cfg(not(any(all(
     target_arch = "arm",
     target_feature = "thumb2",
     target_feature = "v7",
     target_feature = "dsp",
     target_os = "none",
     //target_abi = "eabihf",
-)))]
+), miri)))]
```
JarvisCraft, line 15: Minor nitpicking: I would suggest identing different levels of the epression to make it easier to understand (if Clippy is fine with this):

```suggestion
[cfg(not(any(
    all(
        target_arch = "arm",
        target_feature = "thumb2",
        target_feature = "v7",
        target_feature = "dsp",
        target_os = "none",
        //target_abi = "eabihf",
    ),
    miri
)))]
```

otherwise it is hard to understand to which clause `miri` relates

crates/test/macros/src/lib.rs
```diff
@@ -131,7 +131,7 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
             }
 
             // Test runner entry point
-            fn main(args: *mut u8) -> i32 {
+            pub(super) fn main(args: *mut u8) -> i32 {
```
JarvisCraft, line 134: Since we are now explicitly calling this function, let's mark it `unsafe` describing its `# Safety` invariant (`args` validity requirements)

crates/test/macros/src/lib.rs
```diff
@@ -140,6 +140,17 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
                 }
             }
         }
+
+        #[cfg(all(test, miri))]
+        #[start]
+        fn main(argc: isize, argv: *const *const u8) -> isize {
+            let ret = __test_runner::main(::core::ptr::null_mut());
```
JarvisCraft, line 147: Same here, let's add `unsafe` and the contract for the pointer.

crates/test/macros/src/lib.rs
```diff
@@ -140,6 +140,17 @@ fn tests_runner_impl(args: TokenStream) -> parse::Result<TokenStream> {
...
+
+            // Clean up app state.
+            ::flipperzero_rt::__macro_support::__wait_for_thread_completion();
+
+            ret.try_into().unwrap()
```
JarvisCraft, line 152: Since the only case for failing here is `size_of::<isize>` being `<= size_of::<i32>`, maybe there is a more explicit way to handle this, such as casting (we actually only care about the code being either `0` ot non-`0`, I guess)

JarvisCraft (changes requested):

str4d:
[Rebased on `main`](https://github.com/flipperzero-rs/flipperzero/compare/88ad0d7dc994b9bffcf2dbf39c37c8471251516e..22ee3a3d9c3fd051493dfc8e9c9943d1c5c1d6da).

crates/test/macros/src/lib.rs
str4d: I'd prefer to insulate this unsafety from users entirely by doing the argument-to-string parsing ourselves, as this function in user code is type-constrained by the `entry!` macro. I'll do that in a separate PR.

crates/test/macros/src/lib.rs
str4d: This cannot be `unsafe fn`, as it is required to exactly match the expected type of the `#[start]` function.

str4d:
[Force-pushed](https://github.com/flipperzero-rs/flipperzero/compare/22ee3a3d9c3fd051493dfc8e9c9943d1c5c1d6da..033173a0f4c14ea343d305258f9f372b429f1084) to address comments.

crates/test/macros/src/lib.rs
str4d: #125

dcoles (approved):

JarvisCraft (approved):

JarvisCraft (changes requested):

str4d:
[Rebased on current `main`](https://github.com/flipperzero-rs/flipperzero/compare/94a103596bcbf2ec503fa1f60102651cddd6c913..8c7c40e35145a7c8ca4b7c33a227460476dc9650).

str4d:
Pushed a commit to switch to latest nightly for stabilised `cfg_target_abi` (and fix resulting Clippy lints).

PR #124, opened 2023-12-24
Update firmware to `0.97.1`.

JarvisCraft:
# Description

This updates SDK version to `0.97.1`.

# Removed APIs

`md5` and `sha256` APIs have been removed from the public APIs in flipperdevices/flipperzero-firmware#3234 so it seems that now we should remove them.

I am unaware if there are plans to bring them back in the API form so, as for now, I am hiding our (now broken) bindings to them behind a non-existent feature flag.

# Enhanced Dockerfile

This reworks some aspects of the `Dockerfile` to utilize its caching mechanisms.

I've temporarily disabled running build of `crates/` in it since it blocks us from fixing the API changes whenever there is a build failure (i.e., we fail on tests before producing a new `bindings.rs` required to fix the issue).
This will be a common situation until the API of Flipper is stabilized in `1.0`.


str4d (commented):
utACK 4e631f96cf3a02586a9874a7e3477ab5ce4b0dfb

str4d (approved):

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Migrated to SDK API 35.0 (firmware 0.89.0).
 - `flipperzero_test::tests` now allows `#[cfg(..)]` attributes on test methods.
 
+### Removed
```
str4d, line 32: Oops, I missed that this section was added to a previous release. I'll fix it in my next PR.

PR #125, opened 2023-12-27
rt: Change type signature of entry point to be safe

str4d:
The argument C string pointer is now handled internally instead of in user code.

### DIFF
--- a/crates/rt/src/lib.rs
+++ b/crates/rt/src/lib.rs
@@ -11,6 +11,7 @@ pub mod panic_handler;
 mod thread;
 
 /// The C entry point.
+///
 /// This just delegates to the user's Rust entry point.
 ///
 /// # Safety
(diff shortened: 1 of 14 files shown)


crates/rt/src/lib.rs
```diff
@@ -35,7 +37,15 @@ macro_rules! entry {
         #[export_name = "main"]
         pub unsafe fn __main(args: *mut u8) -> i32 {
             // type check the entry function
-            let f: fn(*mut u8) -> i32 = $path;
+            let f: fn(Option<&::core::ffi::CStr>) -> i32 = $path;
```
dcoles, line 40: This seems like the right direction to take.

Originally I went with `*mut u8` as most of the example apps have a `void* p` argument for main, but the [loader always passes a null-terminated `const char*` (or `NULL`)](https://sourcegraph.com/github.com/flipperdevices/flipperzero-firmware@531ba24e9a1af8ff007e72dd976cb66259895bc7/-/blob/applications/services/loader/loader.c?L321).

dcoles (approved):
lgtm 👍

crates/rt/src/lib.rs
str4d: I expect if they ever change that, it would be accompanied with an SDK version bump (either intentionally or via some other contemporaneous feature), so we can handle it that way if necessary in future.

PR #140, opened 2024-04-02
Fix clippy lints

str4d:


tools/src/bin/fap-lld/main.rs
```diff
@@ -18,6 +18,18 @@ enum Error {
     ObjcopyFailed,
 }
 
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::Io(e) => write!(f, "IO error: {}", e),
+            Error::Parse(e) => write!(f, "Error while parsing ELF: {}", e),
```
dcoles, line 25: You can now put simple variable inline in format strings. For example:
```rust
write!(f, "IO error: {e}")
```

dcoles (approved):

tools/src/bin/fap-lld/main.rs
```diff
@@ -18,6 +18,18 @@ enum Error {
     ObjcopyFailed,
 }
 
+impl fmt::Display for Error {
```
JarvisCraft, line 21: I suggest using `thiserror` here since it would correctly place `source`s and significantly reduce the amount of boilerplate.

JarvisCraft (changes requested):
Generally, looks good to me but I suggest using `thiserror` here.

tools/src/bin/fap-lld/main.rs
str4d: That's orthogonal to this change, which was to silence the clippy lint that the internal errors were not used. And in the tooling here, all that matters is the `Display` impl IIRC.

str4d: Note that I'm not even implementing `std::error::Error`; sourcing etc is irrelevant in this case, as this code is not for a library (so the more appropriate dependency here if any would be `anyhow`).

JarvisCraft: Good point

JarvisCraft (approved):

PR #151, opened 2024-08-11
build(nix): configure build environment via flake

JarvisCraft:
# Description

This PR brings [Nix] support by adding a [`flake.nix`] file.

While not required, this simplifies development on Nix-based systems (NixOS and other ones using Nix package manager).

# Meaning

Since Nix syntax may be new to many, I will explain the file (with its current contents).

What it does is:
* Declares that Rust version should be taken from the `crates/rust-toolchain.toml` file.
* Declares a default development shell (`devShells.default`) in which there `rust` (of the specified version) and `python3` will be installed.

The shell is available via `nix develop` purely (i.e. without anything else from the system) or impurely (i.e., the packages are added on top of the system). `.envrc` file allows the automatic creation of this environment for the user in this directory (being no-op if the user is not using Nix).

# Purpose

While currently this is only to simplify my local development :pekaface:, I intend to use it for some more automations in the repository. For example, I would like to add another dev shell (other from `default`) which would contain [act] for easo of testing GitHub Actions locally.

[Nix]: https://nixos.org/
[`flake.nix`]: https://nixos.wiki/wiki/Flakes
[act]: https://github.com/nektos/act


JarvisCraft:
Merging this as its child #152 has been approved by @dcoles.

PR #155, opened 2024-08-17
Fix for bump SDK to 0.105.0

Shaddy:
I've found the new branch requires updating the runtime with the new API. Submitting this just in case it helps.

Feel free to delete, inspire, whatever. 

Thanks guys!


JarvisCraft:
Thanks for your contribution and the assistance withcthe update!
I've changed the base branch to the one in which `0.105.0` is going to land.

I will take a look at the changes a bit later

crates/rt/src/thread.rs
```diff
@@ -1,7 +1,6 @@
 use core::ffi::CStr;
-use core::ptr;
 
-use flipperzero_sys as sys;
+use flipperzero_sys::{self as sys};
```
JarvisCraft, line 3: This looks like a spurious diff, the previous variant is the preferred one

crates/rt/src/thread.rs
```diff
@@ -16,25 +15,30 @@ pub fn wait_for_completion() {
         );
     }
 
-    const MAX_THREADS: usize = 32;
     let cur_thread_id = unsafe { sys::furi_thread_get_current_id() };
     let app_id = unsafe { CStr::from_ptr(sys::furi_thread_get_appid(cur_thread_id)) };
-    let mut thread_ids: [sys::FuriThreadId; MAX_THREADS] = [ptr::null_mut(); MAX_THREADS];
+    let furi_thread_list = unsafe { sys::furi_thread_list_alloc() };
```
JarvisCraft, line 20: This looks to me as a leak since the allocation is not accompanied by a call to `furi_thread_list_free`.

Worth mentioning that to make it [panic-safe] it would be right to wrap this into a struct with the specific `Drop` implementation (rather than just call `*_free` at the end of the scope).

[panic-safe]: https://doc.rust-lang.org/nomicon/exception-safety.html

crates/rt/src/thread.rs
```diff
@@ -16,25 +15,30 @@ pub fn wait_for_completion() {
...
 
     'outer: loop {
-        let thread_count =
-            unsafe { sys::furi_thread_enumerate(thread_ids.as_mut_ptr(), MAX_THREADS as u32) }
-                as usize;
+        let thread_count = unsafe { sys::furi_thread_list_size(furi_thread_list) };
```
JarvisCraft, line 23: As I can [see][1], the call to `furi_thread_enumerate` is still required to populate the list of threads.

[1]: https://github.com/flipperdevices/flipperzero-firmware/blob/dev/applications/services/cli/cli_commands.c#L396

crates/rt/src/thread.rs
```diff
@@ -16,25 +15,30 @@ pub fn wait_for_completion() {
...
-            let thread_app_id = unsafe { CStr::from_ptr(sys::furi_thread_get_appid(thread_id)) };
+        for thread_index in 0..thread_count {
+            let thread = unsafe { sys::furi_thread_list_get_at(furi_thread_list, thread_index) };
+
+            if thread.is_null() {
+                break;
+            }
+
+            let thread_item = unsafe { *(thread) };
+            let thread_app_id = unsafe { CStr::from_ptr(thread_item.app_id) };
```
JarvisCraft, line 33: Given that this conversion is not no-op (it is $O(n)$ with $n$ being the length of the string) I would suggest moving this evaluation after the evaluation and check of `thread_id` so that you can early `continue` by the latter without computing the value of `thread_app_id`.

JarvisCraft (changes requested):
Here are the things I consider necessary to be changed.

I will probably do these myself, but just posting them here so that I can checklist by them :)

crates/rt/src/thread.rs
Shaddy: For some reason this was automatically created, I thought I removed it. Thanks for pointing out!

crates/rt/src/thread.rs
Shaddy: Absolutely, I've updated with few wrappers, let's discuss if you find it a better solution.

crates/rt/src/thread.rs
Shaddy: I missed this one as well :/ 

Sorry I'm giving more work than I'm solving haha. Thanks for taking the time to review :)

crates/rt/src/thread.rs
Shaddy: Same, with the latest wrappers they are built inline so hope that's a better solution.

Shaddy:
@JarvisCraft I've updated with your suggestions here [78cf788](https://github.com/flipperzero-rs/flipperzero/pull/155/commits/78cf78881cb8951a3f4b68d39d1b8ad9ecb9f8bc) 

It's just a "draft" because I'm not sure if you prefer to use native API directly or you're okay by wrapping stuff.

I've created `ThreadItem`, `ThreadItemList` and a `ThreadListIterator` to deal with the `wait_for_completion` needs.

I would probably create a thin layer that makes the API more friendly, but as said this is personal preference and up to you in this case.

The only thing I wasn't sure, is to add `load` function directly within the `iter()`, as it seems to be tied somehow. For the moment I kept it separated.

Thanks for your review and sorry for the hassle!

JarvisCraft:
> @JarvisCraft I've updated with your suggestions here [78cf788](https://github.com/flipperzero-rs/flipperzero/pull/155/commits/78cf78881cb8951a3f4b68d39d1b8ad9ecb9f8bc)

Thanks!

> It's just a "draft" because I'm not sure if you prefer to use native API directly or you're okay by wrapping stuff.

Generally, it depends on whether the API can be soundly exposed t downstream users, but IMO this is mostly this case.

> I've created `ThreadItem`, `ThreadItemList` and a `ThreadListIterator` to deal with the `wait_for_completion` needs.

Great! Actually, I've also started some local experiments with it and they generally took the same direction (even `ThreadListIterator` looks the same so I guess we are at the right direction).

> I would probably create a thin layer that makes the API more friendly, but as said this is personal preference and up to you in this case.

I am generally for creating such a wrapper, although additional care should be taken to guarantee its soundness. For instance, `FuriThreadListItem::id` is not always valid so access to its `CStr` wrapper should either be `unsafe` or limited via some lifetime.

You can take a look at what I did here: f259ec45394123fc416e84d510213ae2dab77275 (still WIP and I will definitely have no objections to you taking this into your PR).

> The only thing I wasn't sure, is to add `load` function directly within the `iter()`, as it seems to be tied somehow. For the moment I kept it separated.

I think that **ex**plicitly calling `*enumerate*` is the right approach.

> Thanks for your review and sorry for the hassle!

You are welcome, feel free to `@`-summon me if any questions arise :)

Shaddy:
> Generally, it depends on whether the API can be soundly exposed t downstream users, but IMO this is mostly this case.

<3

> Great! Actually, I've also started some local experiments with it and they generally took the same direction (even `ThreadListIterator` looks the same so I guess we are at the right direction).
> I am generally for creating such a wrapper, although additional care should be taken to guarantee its soundness. For instance, `FuriThreadListItem::id` is not always valid so access to its `CStr` wrapper should either be `unsafe` or limited via some lifetime.
> You can take a look at what I did here: [f259ec4](https://github.com/flipperzero-rs/flipperzero/commit/f259ec45394123fc416e84d510213ae2dab77275) (still WIP and I will definitely have no objections to you taking this into your PR).


Good point. I've just merged your code it's way better than my draft, and implemented the DoubleEnded iterator, with the hope is correct :'D


At least it helped me to learn, thank you :)!

JarvisCraft:
> At least it helped me to learn, thank you :)!

I still have some aspects of unsafe Rust to learn myself so take it with a grain of salt (I still have to check some things there), but I guess it may be a good helper)

JarvisCraft:
Thanks for the contribution! The initial change (with some fixes) has been integrated as part of #158.

I will probably try to re-implement the safe wrappers in the future

Issue #150, opened 2024-08-10
Error: Outdated App - SDK Upgrade


Maddox-Werts:
### Hello, Everyone!

I've just updated my Flipper Zero to the latest version of Momentum Firmware. I've also gone ahead and setup a Docker environment to build the application and get everything in order to create the *.FAP* file.

After some Trial and Error, I got Rust to compile to a *.FAP* file. I placed the file on my Flipper Zero's *SD Card* and launched it to get the following message:

![Screenshot-20240810-005302](https://github.com/user-attachments/assets/be3b2b20-9f7e-48c2-8513-f7afac57d523)

Does anyone know how to upgrade the SDK version? Any help would be greatly appreciated.

JarvisCraft:
Hi there and sorry for the late reply!

We are working on update to the latest API in #154 (regenerated bindings) and #155 (additional changes needed in our runtime implementation).

bkstephen:
Any update on this?

Maddox-Werts:
Hey, Bkstephen! 
No, unfortunately not. I was not able to compile it to an updated version of the SDK. I had to write my app in C. I'd recommend you do the same even though Rust is amazing.

bkstephen:
Ok thank you, maybe its worth archiving this project then.

str4d:
No this project will not get archived. We just haven't been able to keep up with the very rapid rate of changes to the Flipper Zero SDK prior to its 1.0 release, so we only generally got updates out that jumped over multiple SDK updates.

[Firmware 1.0 was released today](https://blog.flipper.net/released-firmware-1/), so we will target that for the next update.

ASoldo:
This is what I've been waiting for. No rush, just make it happen eventually. I can't wait to get back to it again!

8de2fdb0:
@str4d 

if you can't wait use
```toml
[dependencies]
flipperzero = { git = "https://github.com/flipperzero-rs/flipperzero", branch = "update-bindings-1.0" }
flipperzero-sys = {  git = "https://github.com/flipperzero-rs/flipperzero", branch = "update-bindings-1.0"  }
flipperzero-rt = {  git = "https://github.com/flipperzero-rs/flipperzero", branch = "update-bindings-1.0"  }
```

@ASoldo nice work

dcoles:
[...]

PR #157, opened 2024-09-11
Update SDK bindings to 73.0 (Firmware 1.0.1)

dcoles:
All the examples appear to be working and the image example is no longer crashing.

Unit test all pass on my updated Flipper running 1.0.1 firmware.

Changes:
- Update SDK bindings to 73.0
- Switch to `unstable-2024-09-10` toolchain (i.e. 1.83.0-nightly)
- Remove `toolbox::md5` and `toolbox::sha256` that are no longer exported by the SDK

Fixes: #150


crates/rt/src/thread.rs
```diff
@@ -16,18 +15,18 @@ pub fn wait_for_completion() {
...
 
-        for &thread_id in thread_ids[..thread_count].iter() {
-            let thread_app_id = unsafe { CStr::from_ptr(sys::furi_thread_get_appid(thread_id)) };
+    'outer: loop {
+        for i in 0..n_threads {
+            let item = unsafe { furi_thread_list_get_at(thread_list, i) };
+            let thread_id = unsafe { furi_thread_get_id((*item).thread) };
+            let thread_app_id = unsafe { CStr::from_ptr((*item).app_id) };
 
             if thread_id == cur_thread_id || thread_app_id != app_id {
```
JarvisCraft, line 31: I suggest that we only lazily initialize `thread_app_id` if the first check (`thread_id == cur_thread_id`) lets us pass.

This is important since `CStr::from_ptr(..)` is $O(n)$

crates/rust-toolchain.toml
JarvisCraft, line 1: Now that both tools and API are built on stable, maybe we could share `rust-toolchain` between them. Especially, considering that `targets` allows multiple values

crates/sys/src/bindings.rs
JarvisCraft, line 1: What do you think of regenerating the commit in which `bindings.rs` are generated with the new CI task? This would allow people exploring the code to validate its provenance (i.e. make sure that nothing suspicious was added to it sine the commit has a CI job associated with it and is committed on behalf of GHA bot)?

JarvisCraft (changes requested):
Great job! Just adding a few notes here :)

crates/Cargo.toml
```diff
@@ -1,19 +1,12 @@
...
-]
+members = ["alloc", "flipperzero", "sys", "rt", "test", "test/macros"]
 resolver = "2"
 
 [workspace.package]
 version = "0.11.0"
 description = "Rust for Flipper Zero"
 edition = "2021"
-rust-version = "1.70.0"
+rust-version = "1.81.0"
```
str4d, line 9: Is this the required MSRV to be able to build with stable Rust?

crates/sys/src/lib.rs
```diff
@@ -5,14 +5,7 @@
...
 // Features that identify thumbv7em-none-eabihf.
 #[cfg(not(any(
-    all(
-        target_arch = "arm",
-        target_feature = "thumb2",
-        target_feature = "v7",
-        target_feature = "dsp",
-        target_os = "none",
-        target_abi = "eabihf",
-    ),
```
str4d, line 15: I would prefer that we comment out the unstable components, like we were doing with `target_abi` before:
```suggestion
// Features that identify thumbv7em-none-eabihf.
// Until target_feature is stable, this also permits TODO.
#[cfg(not(any(
    all(
        target_arch = "arm",
        //target_feature = "thumb2",
        //target_feature = "v7",
        //target_feature = "dsp",
        target_os = "none",
        target_abi = "eabihf",
    ),
```

crates/.cargo/config.toml
```diff
@@ -1,27 +1,31 @@
 # Equivalent to [target.thumbv7em-none-eabihf] but also disabing the runner for Miri.
```
str4d, line 1:
```suggestion
# Equivalent to [target.thumbv7em-none-eabihf] but also disabing the runner for Miri.
# Until target_feature is stable, this also permits TODO.
```
(replacing `TODO` with the other platforms this permits, if any.)

crates/rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2024-04-17"
-targets = [ "thumbv7em-none-eabihf" ]
-components = [ "clippy", "rustfmt" ]
+channel = "stable"
```
str4d, line 2: Is there a benefit to using MSRV here? Probably less so for embedded than for other kinds of library crates, but if we do want to test against MSRV then we'll want to pin it somewhere.

str4d (approved):
utACK with questions and suggestions

crates/Cargo.toml
JarvisCraft: It is, at least since `error_in_core` was stabilized in it.

crates/rust-toolchain.toml
8de2fdb0: It still needs `nighlty` for the cargo feature `different-binary-name`, any idea how to get around that one?

crates/rt/src/thread.rs
dcoles: Good idea. Moved into two separate checks.

crates/sys/src/lib.rs
dcoles: Sure. Added this back in.

crates/sys/src/bindings.rs
dcoles: Running the workflow now: https://github.com/flipperzero-rs/flipperzero/actions/runs/10823803133

crates/rust-toolchain.toml
dcoles: That's unfortunate. While `different-binary-name` isn't strictly necessary, it does make for a much better overall experience. Let me see if I can think of something.

dcoles: We might be able to do that from `fap-lld`. Let me see how hard that change would be.

crates/sys/src/bindings.rs
dcoles: Rebased on #158.

crates/rust-toolchain.toml
dcoles: So while yes, `fap-lld` can rename the file, this is too early in the build process. At this stage the binary is still in `{name}-{hash}` form before being renamed to the final binary.

So I don't think there's any way short of a wrapper script or sticking with nightly.

dcoles: Reverted the `stable` change and switched to `nightly-2024-09-10` instead.

crates/rust-toolchain.toml
dcoles: Seems we're sticking with nightly for the moment.

JarvisCraft (approved):

JarvisCraft:
@dcoles, could you please write the release notes for the release in GitHub Releases page?

Please note, that `main` is now beyond the release commit, so the one you may want to tag is probably 2b1c52b577a234be7827442607583a53efda9ddd.

dcoles:
> @dcoles, could you please write the release notes for the release in GitHub Releases page?

Thanks for the reminder. I've pointed it at the [`v0.12.0`](https://github.com/flipperzero-rs/flipperzero/releases/tag/v0.12.0) tag.

PR #170, opened 2024-10-08
Initial SubGhz development

QuantumEF:
Initial work, opening draft pull request to solicit feedback and comments.

This is not yet finished as I have not yet implemented the builder for the CustomSubGhzPreset to allow for custom presets.

### DIFF
--- a/crates/flipperzero/Cargo.toml
+++ b/crates/flipperzero/Cargo.toml
@@ -41,6 +41,7 @@ embedded-hal-0 = { package = "embedded-hal", version = "0.2.7", features = ["unp
 
 # Docs
 document-features = { workspace = true, optional = true }
+uom = { version = "0.36.0", features = ["si", "u32"], default-features = false }
 
 [dev-dependencies]
 flipperzero-alloc.workspace = true
--- a/crates/flipperzero/src/subghz/device.rs
+++ b/crates/flipperzero/src/subghz/device.rs
@@ -1,0 +1,199 @@
+//! Safe wrapper utilizing the API provided in `lib/subghz/devices/devices.h`
+
+use super::{error::SubGhzError, preset::SubGhzBuiltinPreset};
+use core::{ffi::CStr, ptr::null_mut};
+use flipperzero_sys::{
+    subghz_devices_begin, subghz_devices_end, subghz_devices_flush_rx, subghz_devices_flush_tx,
+    subghz_devices_get_by_name, subghz_devices_idle, subghz_devices_is_connect,
+    subghz_devices_is_frequency_valid, subghz_devices_is_rx_data_crc_valid,
+    subghz_devices_load_preset, subghz_devices_read_packet, subghz_devices_reset,
+    subghz_devices_rx_pipe_not_empty, subghz_devices_set_frequency, subghz_devices_set_rx,
+    subghz_devices_set_tx, subghz_devices_sleep, subghz_devices_write_packet, SubGhzDevice,
+};
+use uom::si::{frequency::hertz, u32::Frequency};
+
+#[cfg(feature = "alloc")]
+use super::preset::SubGhzPreset;
+
+pub struct SubGhz {
+    device: &'static SubGhzDevice,
+}
+
+impl SubGhz {
+    pub fn subghz_devices_get_by_name(name: &CStr) -> Option<SubGhz> {
+        // Safety: input type enforeced by CStr type.
+        let dev = unsafe { subghz_devices_get_by_name(name.as_ptr()) };
+        // Safety: This pointer should either be null or return a reference, I don't see how it could not.
+        // Further, the reference should be static as it relates to the variable:
+        // `static SubGhzDeviceRegistry* subghz_device_registry = NULL;` in `lib/subghz/devices/registry.c``
+        let dev = unsafe { dev.as_ref()? };
+        Some(Self { device: dev })
+    }
+
+    // No clue what sort of error may occur, if using the internal device, it will always return Ok()
+    pub fn begin(&mut self) -> Result<(), SubGhzError> {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        let result = unsafe { subghz_devices_begin(self.device) };
+
+        // False indicates Ok
+        if result {
+            Err(SubGhzError::UnableToOpenDevice)
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn is_connect(&mut self) -> bool {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        unsafe { subghz_devices_is_connect(self.device) }
+    }
+
+    pub fn reset(&mut self) {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        unsafe {
+            subghz_devices_reset(self.device);
+        }
+    }
+
+    pub fn sleep(&mut self) {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        unsafe {
+            subghz_devices_sleep(self.device);
+        }
+    }
+
+    pub fn idle(&mut self) {
+        unsafe {
+            subghz_devices_idle(self.device);
+        }
+    }
+
+    #[cfg(feature = "alloc")]
+    pub fn load_preset(&mut self, preset: SubGhzPreset) {
+        let furi_preset = preset.into_furi_preset();
+
+        let preset_data = if let SubGhzPreset::Custom(data) = preset {
+            data.as_mut_ptr()
+        } else {
+            null_mut::<u8>()
+        };
+
+        // Safety: Type enforcement of the builtin presets, so a valid pointer is passed based on the input Preset.
+        unsafe {
+            subghz_devices_load_preset(self.device, furi_preset, preset_data);
+        }
+    }
+
+    pub fn load_builtin_preset(&mut self, preset: SubGhzBuiltinPreset) {
+        let furi_preset = preset.into_furi_preset();
+
+        // Safety: Type enforcement of the builtin presets, so it is Ok to pass in the null pointer here.
+        unsafe {
+            subghz_devices_load_preset(self.device, furi_preset, null_mut::<u8>());
+        }
+    }
+
+    // If the returned frequency from the internal API function call is 0, this function returns an error.
+    pub fn set_frequency(&mut self, freq: Frequency) -> Result<Frequency, SubGhzError> {
+        let freq_u32 = freq.value;
+        let actual_freq = unsafe { subghz_devices_set_frequency(self.device, freq_u32) };
+
+        if actual_freq != 0 {
+            Ok(Frequency::new::<hertz>(actual_freq))
+        } else {
+            Err(SubGhzError::UnableToSetFrequency)
+        }
+    }
+
+    pub fn is_frequency_valid(&mut self, freq: Frequency) -> bool {
+        unsafe { subghz_devices_is_frequency_valid(self.device, freq.value) }
+    }
+
+    pub fn set_async_mirror_pin(&mut self) {
+        unimplemented!("Should probably create safe abstraction around GpioPin first, have not investigated if this is needed");
+    }
+
+    pub fn get_data_gpio(&mut self) {
+        unimplemented!("Should probably create safe abstraction around GpioPin first, have not investigated if this is needed");
+    }
+
+    pub fn set_tx(&mut self) -> Result<(), SubGhzError> {
+        let result = unsafe { subghz_devices_set_tx(self.device) };
+
+        // Returns true if Ok
+        if result {
+            Ok(())
+        } else {
+            Err(SubGhzError::CannotTxOnFrequency)
+        }
+    }
+
+    pub fn flush_tx(&mut self) {
+        unsafe { subghz_devices_flush_tx(self.device) };
+    }
+
+    pub fn set_rx(&mut self) {
+        unsafe {
+            subghz_devices_set_rx(self.device);
+        }
+    }
+
+    pub fn flush_rx(&mut self) {
+        unsafe {
+            subghz_devices_flush_rx(self.device);
+        }
+    }
+
+    pub fn get_rssi(&mut self) -> Option<f32> {
+        let rssi_function = unsafe { self.device.interconnect.as_ref() }?.get_rssi?;
+        Some(unsafe { rssi_function() })
+    }
+
+    pub fn get_lqi(&mut self) -> Option<u8> {
+        let lqi_function = unsafe { self.device.interconnect.as_ref()?.get_lqi? };
+        Some(unsafe { lqi_function() })
+    }
+
+    pub fn rx_pipe_not_empty(&mut self) -> bool {
+        unsafe { subghz_devices_rx_pipe_not_empty(self.device) }
+    }
+
+    pub fn is_rx_crc_valid(&mut self) -> bool {
+        unsafe { subghz_devices_is_rx_data_crc_valid(self.device) }
+    }
+
+    pub fn read_packet(&mut self, buf: &mut [u8]) -> usize {
+        let mut size = if buf.len() > 255 {
+            255_u8
+        } else {
+            buf.len() as u8
+        };
+
+        unsafe {
+            subghz_devices_read_packet(self.device, buf.as_mut_ptr(), &mut size);
+        };
+        size as usize
+    }
+
+    pub fn write_packet(&mut self, buf: &[u8]) -> Result<(), SubGhzError> {
+        if buf.len() > 255 {
+            return Err(SubGhzError::PacketTooLong);
+        }
+
+        let size = buf.len() as u8;
+
+        unsafe {
+            subghz_devices_write_packet(self.device, buf.as_ptr(), size);
+        }
+        Ok(())
+    }
+}
+
+impl Drop for SubGhz {
+    fn drop(&mut self) {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        unsafe {
+            subghz_devices_end(self.device);
+        }
+    }
+}
--- a/crates/flipperzero/src/subghz/error.rs
+++ b/crates/flipperzero/src/subghz/error.rs
@@ -1,0 +1,22 @@
+use core::{error::Error, fmt::Display};
+
+#[derive(Clone, Copy, Debug)]
+pub enum SubGhzError {
+    UnableToSetFrequency,
+    PacketTooLong,
+    CannotTxOnFrequency,
+    UnableToOpenDevice,
+}
+
+impl Display for SubGhzError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::UnableToSetFrequency => write!(f, "Unable to set the frequency"),
+            Self::PacketTooLong => write!(f, "The packet length is too long, the device can only be given packets of max length 255"),
+            Self::CannotTxOnFrequency => write!(f,"Unable to Tx on the given frequeny (Region Locked?)"),
+            Self::UnableToOpenDevice => write!(f,"Was unable to get a handle to the device")
+        }
+    }
+}
+
+impl Error for SubGhzError {}
--- a/crates/flipperzero/src/subghz/preset.rs
+++ b/crates/flipperzero/src/subghz/preset.rs
@@ -1,0 +1,134 @@
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
+use flipperzero_sys::{
+    FuriHalSubGhzPreset, FuriHalSubGhzPreset_FuriHalSubGhzPreset2FSKDev238Async,
+    FuriHalSubGhzPreset_FuriHalSubGhzPreset2FSKDev476Async,
+    FuriHalSubGhzPreset_FuriHalSubGhzPresetCustom,
+    FuriHalSubGhzPreset_FuriHalSubGhzPresetGFSK9_99KbAsync,
+    FuriHalSubGhzPreset_FuriHalSubGhzPresetIDLE,
+    FuriHalSubGhzPreset_FuriHalSubGhzPresetMSK99_97KbAsync,
+    FuriHalSubGhzPreset_FuriHalSubGhzPresetOok270Async,
+    FuriHalSubGhzPreset_FuriHalSubGhzPresetOok650Async,
+};
+
+#[cfg(feature = "alloc")]
+#[derive(Clone, Debug)]
+pub enum SubGhzPreset {
+    /// default configuration
+    IDLE,
+    /// OOK, bandwidth 270kHz, asynchronous
+    Ook270Async,
+    /// OOK, bandwidth 650kHz, asynchronous
+    Ook650Async,
+    /// FM, deviation 2.380371 kHz, asynchronous
+    _2FSKDev238Async,
+    /// FM, deviation 47.60742 kHz, asynchronous
+    _2FSKDev476Async,
+    /// MSK, deviation 47.60742 kHz, 99.97Kb/s, asynchronous
+    MSK99_97KbAsync,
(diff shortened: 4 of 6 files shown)


QuantumEF:
With this latest commit, the following basic app does work and does not crash.

```rust
//! Template project for Flipper Zero.
//! This app prints "Hello, Rust!" to the console then exits.

#![no_main]
#![no_std]

// Required for panic handler
// use flipperzero_rt;

extern crate flipperzero_alloc;

use core::ffi::CStr;

use flipperzero::dialogs::alert;
use flipperzero::println;
use flipperzero::subghz;
use flipperzero_rt::{entry, manifest};

// Define the FAP Manifest for this application
manifest!(
    name = "Flipper Zero Rust",
    app_version = 1,
    has_icon = true,
    // See https://github.com/flipperzero-rs/flipperzero/blob/v0.11.0/docs/icons.md for icon format
    icon = "rustacean-10x10.icon",
);

// Define the entry function
entry!(main);

// Entry point
fn main(_args: Option<&CStr>) -> i32 {
    alert("Hello Erik");

    let dev_name = c"cc1101_int";

    let x = subghz::SubGhz::subghz_devices_get_by_name(dev_name);

    match x {
        Some(_) => alert("YEA"),
        None => alert("Boooo"),
    }

    0
}
```

QuantumEF:
@dcoles Do you have time to give feedback on what I have done so far? ie: just how I have started going about doing things.
As mentioned in my previous comment, I have only tested just opening device. I am continuing testing to validate the functionality of other things.

crates/flipperzero/Cargo.toml
```diff
@@ -41,6 +41,7 @@ embedded-hal-0 = { package = "embedded-hal", version = "0.2.7", features = ["unp
 
 # Docs
 document-features = { workspace = true, optional = true }
+uom = { version = "0.36.0", features = ["si", "u32"], default-features = false }
```
JarvisCraft, line 44: I suggest that we add all new dependencies to workspace root and reference them here via `workspace = true`

Also note, that this part of `Cargo.toml` is for docs-related dependencies (see comment two lines above)

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
+//! Safe wrapper utilizing the API provided in `lib/subghz/devices/devices.h`
+
+use super::{error::SubGhzError, preset::SubGhzPreset};
+use core::{ffi::CStr, ptr::null_mut};
+use flipperzero_sys::{
```
JarvisCraft, line 5: I feel like importing every function from `sys` only to call it once complicates this source so I suggest aliasing `flipperzero_sys` as `sys` and using the qualified paths, i.e.:

```rust
use flipperzero_sys as sys;
```

and

```rust
let result = unsafe { sys::subghz_devices_begin(self.device) };
```

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+    subghz_devices_write_packet, SubGhzDevice,
+};
+use uom::si::{frequency::hertz, u32::Frequency};
+
+pub struct SubGhz {
+    device: &'static SubGhzDevice,
+}
+
+impl SubGhz {
+    /// For the internal device use "cc1101_int" (From `lib/subghz/devices/cc1101_int/cc1101_int_interconnect.c`)
```
JarvisCraft, line 21: A link would be useful, I think

Note that there is cool Markdown syntax for link in footers:

```suggestion
    /// For the internal device use [`cc1101_int`]
    ///
    /// [`cc1101_int`]: https://github.com/flipperdevices/flipperzero-firmware/blob/dev/lib/subghz/devices/cc1101_int/cc1101_int_interconnect.c
```

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub fn subghz_devices_get_by_name(name: &CStr) -> Option<SubGhz> {
+        unsafe { subghz_devices_init() }
+
+        // Safety: input type enforeced by CStr type.
```
JarvisCraft, line 25: I am not sure that this is sound: `CStr` only guarantees pointer validity and the absence of `\0`s, but it does *not* guarantee that the string is valid UTF-8 and (but I may be mistaken here) some Furi APIs require more specific strings.

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+        let dev = unsafe { subghz_devices_get_by_name(name.as_ptr()) };
+        // Safety: This pointer should either be null or return a reference, I don't see how it could not.
+        // Further, the reference should be static as it relates to the variable:
+        // `static SubGhzDeviceRegistry* subghz_device_registry = NULL;` in `lib/subghz/devices/registry.c``
+        let dev = unsafe { dev.as_ref()? };
```
JarvisCraft, line 30: Let's move `?` out of `unsafe`

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+        Some(Self { device: dev })
```
JarvisCraft, line 31: Probably worth renaming the variable to `device` to use shorthand form here:

```suggestion
        Some(Self { device })
```

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    // No clue what sort of error may occur, if using the internal device, it will always return Ok()
+    pub fn begin(&mut self) -> Result<(), SubGhzError> {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        let result = unsafe { subghz_devices_begin(self.device) };
```
JarvisCraft, line 37: `failed` seems to be a more easy-to-review name for the variable

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+        // Safety: This pointer should either be null or return a reference, I don't see how it could not.
+        // Further, the reference should be static as it relates to the variable:
+        // `static SubGhzDeviceRegistry* subghz_device_registry = NULL;` in `lib/subghz/devices/registry.c``
+        let dev = unsafe { dev.as_ref()? };
+        Some(Self { device: dev })
+    }
+
+    // No clue what sort of error may occur, if using the internal device, it will always return Ok()
+    pub fn begin(&mut self) -> Result<(), SubGhzError> {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
```
JarvisCraft, line 36: Important: `furi_check` causing a crash is *not* unsafe thus it should not be mentioned in safety comments (in the same matter as `panic!` is safe to call), although mentioning non-nullity is useful.

What should be mentioned here is if there are any `*const SubGhzDevice` invariants which should be upheld.

Same goes for the following `unsafe` calls in the file.

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+    subghz_devices_is_connect, subghz_devices_is_frequency_valid,
+    subghz_devices_is_rx_data_crc_valid, subghz_devices_load_preset, subghz_devices_read_packet,
+    subghz_devices_reset, subghz_devices_rx_pipe_not_empty, subghz_devices_set_frequency,
+    subghz_devices_set_rx, subghz_devices_set_tx, subghz_devices_sleep,
+    subghz_devices_write_packet, SubGhzDevice,
+};
+use uom::si::{frequency::hertz, u32::Frequency};
+
+pub struct SubGhz {
+    device: &'static SubGhzDevice,
```
JarvisCraft, line 17: Storing `*const SubGhzDevice` as a Rust reference is really tricky, considering that there are much more invariant to uphold for it, especially given that `SubGhzDevice` seems to have some form of interior mutability internally. I suggest that you store `NonNull<SubGhzDevice>` instead, only creating short-living references when needed (e.g. when reading fields).

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub fn is_rx_crc_valid(&mut self) -> bool {
+        unsafe { subghz_devices_is_rx_data_crc_valid(self.device) }
+    }
+
+    pub fn read_packet(&mut self, buf: &mut [u8]) -> usize {
+        let mut size = if buf.len() > 255 {
+            255_u8
+        } else {
+            buf.len() as u8
+        };
```
JarvisCraft, line 163:
```suggestion
        let mut size = u8::try_from(buf.len()).unwrap_or(u8::MAX);
```

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+        Ok(())
+    }
+}
+
+impl Drop for SubGhz {
+    fn drop(&mut self) {
+        // Safety: self.device is not Null so this will not crash when furi_check is invoked.
+        unsafe {
+            subghz_devices_end(self.device);
+            subghz_devices_deinit();
```
JarvisCraft, line 190: What happens if multiple instances of `SubGhz` are created and thus `subghz_devices_deinit` is called (on ones destructor) while the other one is still active?

I guess we may need to address this upstream, e.g. request adding a counter to it (e.g. the number of currently active *activations*)

crates/flipperzero/src/subghz/device.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        unsafe {
+            subghz_devices_read_packet(self.device, buf.as_mut_ptr(), &mut size);
+        };
+        size as usize
+    }
+
+    pub fn write_packet(&mut self, buf: &[u8]) -> Result<(), SubGhzError> {
+        if buf.len() > 255 {
+            return Err(SubGhzError::PacketTooLong);
```
JarvisCraft, line 173: I suggest adding `len` field to this error variant

crates/flipperzero/src/subghz/preset.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+/// This struct is defined for creating custom configurations. It enforces the null padding needed at the end. along with the power amplifier config?
+pub struct CustomSubGhzPreset<'a>(&'a [u8]);
+
+impl<'a> CustomSubGhzPreset<'a> {
+    /// # Safety
```
JarvisCraft, line 55:
```suggestion
    /// # Safety
    ///
```


Markdown requires a newline after a header

crates/flipperzero/src/subghz/preset.rs
```diff
@@ -0,0 +1,64 @@
...
+            }
+            SubGhzPreset::MSK99_97KbAsync => FuriHalSubGhzPreset_FuriHalSubGhzPresetMSK99_97KbAsync,
+            SubGhzPreset::GFSK9_99KbAsync => FuriHalSubGhzPreset_FuriHalSubGhzPresetGFSK9_99KbAsync,
+            SubGhzPreset::Custom(_) => FuriHalSubGhzPreset_FuriHalSubGhzPresetCustom,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+/// This struct is defined for creating custom configurations. It enforces the null padding needed at the end. along with the power amplifier config?
```
JarvisCraft, line 51:
```suggestion
/// This struct is defined for creating custom configurations. It enforces the null padding needed at the end. along with the power amplifier config?
#[derive(Clone, Debug)]
```

crates/flipperzero/src/subghz/preset.rs
```diff
@@ -0,0 +1,64 @@
...
+pub struct CustomSubGhzPreset<'a>(&'a [u8]);
+
+impl<'a> CustomSubGhzPreset<'a> {
+    /// # Safety
+    /// Read the manual for the CC1101 chip and review the [`furi_hal_subghz_load_custom_preset()`](https://github.com/flipperdevices/flipperzero-firmware/blob/0eaad8bf64f01a6f932647a9cda5475dd9ea1524/targets/f7/furi_hal/furi_hal_subghz.c#L162) function provided by the flipper API
```
JarvisCraft, line 56:
```suggestion
    /// Read the manual for the CC1101 chip and review the [`furi_hal_subghz_load_custom_preset()`].
    ///
    /// [`furi_hal_subghz_load_custom_preset()`]: (https://github.com/flipperdevices/flipperzero-firmware/blob/0eaad8bf64f01a6f932647a9cda5475dd9ea1524/targets/f7/furi_hal/furi_hal_subghz.c#L162) function provided by the flipper API
```

crates/flipperzero/src/subghz/preset.rs
```diff
@@ -0,0 +1,64 @@
...
+    _2FSKDev476Async,
+    /// MSK, deviation 47.60742 kHz, 99.97Kb/s, asynchronous
+    MSK99_97KbAsync,
+    /// GFSK, deviation 19.042969 kHz, 9.996Kb/s, asynchronous
+    GFSK9_99KbAsync,
+    Custom(CustomSubGhzPreset<'a>),
+}
+
+impl<'a> SubGhzPreset<'a> {
+    pub fn into_furi_preset(&self) -> FuriHalSubGhzPreset {
```
JarvisCraft, line 32: Not sure, but maybe its worth adding `impl From<&SubGhzPreset<'_>> for FuriHalSubGhzPreset>`

Also, `as_` prefix [seems to be][1] more valid for the function

[1]: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

crates/flipperzero/src/subghz/preset.rs
```diff
@@ -0,0 +1,64 @@
...
+            SubGhzPreset::Ook650Async => FuriHalSubGhzPreset_FuriHalSubGhzPresetOok650Async,
+            SubGhzPreset::_2FSKDev238Async => {
+                FuriHalSubGhzPreset_FuriHalSubGhzPreset2FSKDev238Async
+            }
+            SubGhzPreset::_2FSKDev476Async => {
+                FuriHalSubGhzPreset_FuriHalSubGhzPreset2FSKDev476Async
+            }
+            SubGhzPreset::MSK99_97KbAsync => FuriHalSubGhzPreset_FuriHalSubGhzPresetMSK99_97KbAsync,
+            SubGhzPreset::GFSK9_99KbAsync => FuriHalSubGhzPreset_FuriHalSubGhzPresetGFSK9_99KbAsync,
+            SubGhzPreset::Custom(_) => FuriHalSubGhzPreset_FuriHalSubGhzPresetCustom,
```
JarvisCraft, line 45:
```suggestion
            Self::IDLE => FuriHalSubGhzPreset_FuriHalSubGhzPresetIDLE,
            Self::Ook270Async => FuriHalSubGhzPreset_FuriHalSubGhzPresetOok270Async,
            Self::Ook650Async => FuriHalSubGhzPreset_FuriHalSubGhzPresetOok650Async,
            Self::_2FSKDev238Async => {
                FuriHalSubGhzPreset_FuriHalSubGhzPreset2FSKDev238Async
            }
            Self::_2FSKDev476Async => {
                FuriHalSubGhzPreset_FuriHalSubGhzPreset2FSKDev476Async
            }
            Self::MSK99_97KbAsync => FuriHalSubGhzPreset_FuriHalSubGhzPresetMSK99_97KbAsync,
            Self::GFSK9_99KbAsync => FuriHalSubGhzPreset_FuriHalSubGhzPresetGFSK9_99KbAsync,
            Self::Custom(_) => FuriHalSubGhzPreset_FuriHalSubGhzPresetCustom,
```

crates/flipperzero/src/subghz/preset.rs
```diff
@@ -0,0 +1,64 @@
+use flipperzero_sys::{
```
JarvisCraft, line 1: Same as preciously, a single `use flipperzero_sys as sys` would look cleaner

crates/flipperzero/src/subghz/error.rs
```diff
@@ -0,0 +1,22 @@
+use core::{error::Error, fmt::Display};
+
+#[derive(Clone, Copy, Debug)]
```
JarvisCraft, line 3: `uDebug` should also be implemented when `Debug` is implemented.

JarvisCraft (changes requested):
The first iteration of the changes

They maybe nitpicky, but I wish that they will be helpful

crates/flipperzero/Cargo.toml
QuantumEF: Makes sense to me, I will hold off on this change for now since I need to properly figure out how to use uom crate since at present my use of it utilizes dynamic dispatch. I also want to read up on cargo workspaces since I have not used them before.

crates/flipperzero/src/subghz/device.rs
QuantumEF: Based on what I saw, the string is eventually used by the following function: https://github.com/flipperdevices/flipperzero-firmware/blob/release/lib/subghz/registry.c#L3

My interpretation is that I just needed to make sure I had a null terminated array as the input. That being said, I was looking at commit 10 behind that I linked to, so I should probably look again since it looks like the file changed a fair bit.

crates/flipperzero/src/subghz/preset.rs
QuantumEF: I am probably going to hold off for now on implementing this since [the docs](https://doc.rust-lang.org/std/convert/trait.From.html) mention that "The From trait is intended for perfect conversions," This function looses data when converting from the `Custom` variant.

I like the `as_` suggestion because it is something I always forget the semantics about (I should print it out, lol).
When I originally created some of the structs, I was not able to implement copy, but now I am  since the custom variant is `Custom(&[u8])`. Because of this, I am wondering if I should just change the function signature to take in `self` instead of a `&self` then maybe implying is should use `to_` ?

crates/flipperzero/src/subghz/preset.rs
QuantumEF: Changed with 37c523b869a3a454792d2ff8c6b23d5d0c6f6f31

crates/flipperzero/src/subghz/error.rs
QuantumEF: I didn't realize this, added with 1ebd7ca9132f77ced2fcc554628f39cdac56c5d6

crates/flipperzero/src/subghz/device.rs
QuantumEF: In some haphazard test code:

```rust
fn main(_args: Option<&CStr>) -> i32 {
    let dev_name = c"cc1101_int";

    let Some(mut x) = subghz::SubGhz::subghz_devices_get_by_name(dev_name) else {
        alert("BOOOO");
        return 0;
    };

    x.load_preset(subghz::preset::SubGhzPreset::Ook270Async);

    let Ok(()) = x.begin() else {
        return 0;
    };

    let Ok(actual_freq) = x.set_frequency(Frequency::new::<hertz>(433_500_000)) else {
        return 0;
    };

    let mut s = String::<128>::new();

    alert("about to take second ref");
    // Crash
    let Some(y) = subghz::SubGhz::subghz_devices_get_by_name(dev_name) else {
        alert("Cant take second reference");
        return 0;
    };

    0
}
```

Attempting to create a second reference crashes the device (I am not too surprised since the API does a lot of checking for safety). That being said, I don't know which function actually crashes things, so I don't know how it is enforced: I would like to be able to enforce this to prevent the crash.

crates/flipperzero/src/subghz/preset.rs
QuantumEF: Changed with a3036929a2a702be8313d927819880dd94544a85
(Was not able to use the "Commit Suggestion" feature since the change is "outdated" due to all the other changes I have committed as I have been going through your suggestions.)

crates/flipperzero/src/subghz/preset.rs
QuantumEF: Not quite sure what exactly the change is here. My assumption is for it to include both a link to the function in the `flipperzero_sys::furi_hal_etc` documentation as well as the link I provide to the C source code?

crates/flipperzero/src/subghz/device.rs
QuantumEF: I am aware that a crash is considered safe; as you mentioned I wanted to put it there to mention the non-nullity.

This is something I want to look into more, so thanks for your feedback on this. I will mark this comment as resolved as you have more details in your other comment.

QuantumEF: As a sidenote, I think I lost a bunch of replys to some of your comments since I clicked "Resolve Conversation" without adding the comment :/ I will go back and try add those comments, but most of them were just links to the commits in which I fixed them.

QuantumEF:
> The first iteration of the changes
> 
> They maybe nitpicky, but I wish that they will be helpful

Thank you so much for your feedback and help! As you see I have gone through most of your suggestions and added them. The ones I have not marked as resolved yet are because I did want to think on them and possible discuss them more. 

In my error, I lost some of the comments on the changes I marked as resolved because I assumed the "Marked as resolved" button did something similar to the "Comment and mark as resolved" button I have seen elsewhere; Oops, most of those comments were just referencing the commit I changed things in.

crates/flipperzero/src/subghz/device.rs
QuantumEF: Good idea. I should also probably do this with the other variants as well. For now I will just do it for the PacketTooLong variant

crates/flipperzero/src/subghz/device.rs
QuantumEF: To add additional info, the string is compared with strcmp() to find the matching device in the list.

And no, CStr does not guarantee valid UTF-8 (it can be created from any &[u8] and the "constructors" only are checking for null termination.)

JarvisCraft:
> Thank you so much for your feedback and help!

You are welcome! Thank you for working on this feature!

> As you see I have gone through most of your suggestions and added them. The ones I have not marked as resolved yet are because I did want to think on them and possible discuss them more.

For sure! I understand that this is a draft, so my review should be considered as more of the *recommendations* by now :)

> In my error, I lost some of the comments on the changes I marked as resolved because I assumed the "Marked as resolved" button did something similar to the "Comment and mark as resolved" button I have seen elsewhere; Oops, most of those comments were just referencing the commit I changed things in.

No problems, GitHub is a bit weird with PR review threads anyway

JarvisCraft:
It may take me some time to continue the review (I will be at the conference for a few days) thus feel free to leave some questions unresolved as-is since eventually the will be reviewed

crates/flipperzero/src/subghz/preset.rs
JarvisCraft: My bad, I've accidentally trimmed the end of the comment sentence.

The intention here is to move the URL to the markdown footer so that the header text is easier to read

crates/flipperzero/src/subghz/device.rs
JarvisCraft: Nice, given that it is only used for `strcmp`, `CStr` fits the purpose

crates/flipperzero/src/subghz/device.rs
QuantumEF: So looking at [this instance of a SubGhzDevice](https://github.com/flipperdevices/flipperzero-firmware/blob/fe424061c968c41131448697ea587533402dccae/lib/subghz/devices/cc1101_int/cc1101_int_interconnect.c#L93), it looks like a given SubGhzDevice is going to be a struct that contains a name and a pointer to a struct that contains a bunch of function pointers, so I don't think this is an issue; but I also do not have much practical experience writing in C, so I could not be interpreting things correctly. All that being said, making the change back to `*const SubGhzDevice` is simple and only effect the `get_rssi` ang `get_lqi` functions, though I am not sure how should now change the semantics of those functions: ie should they still be `Option<T>` or `Result<T>` to indicate if the device pointer is null or if the function pointer is null.

PR #173, opened 2024-10-13
Generate bindings at compile-time

dcoles:
This replaces the `generate-bindings` script with a `build.rs` script for `flipperzero-sys` that downloads the latest SDK and uses that to generate Rust bindings.

Fixes: #41


dcoles:
Uploading for initial review and testing.

JarvisCraft:
I think that we should provide a way to build the crate offline from scratch.
Some build systems and distros (for instance, `nixpkgs` of NixOS) are quite  restrictive towards internet access during the build phase and thus may be unlucky trying to build the crate depending on internet download during `build.rs` evaluation.

What I see as the requirement for this PR is a way to have:
1. Some default source of truth available with the crate itself (i.e. not requiring internet access during build).
2. *Opt-in* ability to override via local path (this can actually be enough to not implement point 1).
3. *opt-in* ability to override it either via upstream flipperzero repo download (what you do in this PR);

So ideally I would like to see either:

| Mode | Default | Envs (alternatively, a crate feature) | Behavior | Works offline |
|---|---|---|---|---|
| Pre-built | :white_check_mark: | None of the following | Uses something predefined, e.g. our pre-generated `bindings.rs`, as done now | :white_check_mark: |
| Local path | :no_entry: / :white_check_mark: (if we decide to get rid of Pre-built) | `FLIPPER_SDK_PATH` | Generates bindings from the local SDK installation | :white_check_mark: |
| Download | :no_entry: | `FLIPPER_SDK_VERSION`, `FLIPPER_TOOLCHAIN_VERSION`, maybe explicit internet access opt-in flag | Downloads the SDK by the given coordinates (by the way, I suggest adding the ability to override `BUILDS_URL` for the sake of custom SDKs / forks) | :no_entry: |

PR #179, opened 2024-10-21
Standardize use of `furi::time::Duration`

dcoles:
Most places use `furi::time::Duration` for things like timeouts, though there are a few place that used `core::time::Duration` (typically before `furi::time::Duration`).

I left `furi::thread::sleep` unchanged but introduced a new `sleep_ticks` to specifically support `furi::time::Duration`.


crates/flipperzero/src/furi/thread.rs
```diff
@@ -203,6 +205,15 @@ pub fn sleep(duration: core::time::Duration) {
     }
 }
 
+/// Puts the current thread to sleep for at least the specified amount of time.
+///
+/// This differs from [`sleep`] in that a Furi [`Duration`] can represent an exact number of ticks.
+pub fn sleep_ticks(duration: Duration) {
```
JarvisCraft, line 211: Are we more interested in users using this instead of `sleep`? If yes then maybe we want it to have an easier `sleep` name with `sleep_<smth>` be given to the former.

Also, it would be useful to also give a reference to this function from the former's doc.

crates/flipperzero/src/furi/time.rs
```diff
@@ -464,9 +470,36 @@ impl<'a> Sum<&'a Duration> for Duration {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, uDebug)]
+pub struct TryFromDurationError;
+
+impl core::error::Error for TryFromDurationError {}
+
+impl core::fmt::Display for TryFromDurationError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
```
JarvisCraft, line 479: Nit: I suggest importing `core::fmt` in scope

JarvisCraft (changes requested):

crates/flipperzero/src/furi/time.rs
dcoles: Fixed. I did the same for the other `core::` uses.

crates/flipperzero/src/furi/thread.rs
dcoles: I'm a bit torn on this one. Originally the plan was to make `sleep` take a `furi::time::Duration`, but `core::time::Duration` can support sub-tick (typically 1 ms) durations which can be used with `furi_delay_ms`.

Updated the documentation to explain the difference between the two.

JarvisCraft (approved):

PR #180, opened 2024-10-21
Add `flipperzero::furi::kernel` module

dcoles:
This module exposes most of the `furi_kernel_*` APIs.

Centralizing these lets us remove a number of `unsafe` calls peppered across the codebase and allows things like providing optimized versions.

### DIFF
--- a/crates/flipperzero/src/furi/kernel.rs
+++ b/crates/flipperzero/src/furi/kernel.rs
@@ -1,0 +1,90 @@
+//! Furi Kernel primitives.
+
+use flipperzero_sys as sys;
+use ufmt::derive::uDebug;
+
+use crate::furi;
+
+/// Check if CPU is in IRQ; or kernel running and IRQ is masked.
+pub fn is_irq_or_masked() -> bool {
+    unsafe { sys::furi_kernel_is_irq_or_masked() }
+}
+
+/// Check if kernel is running.
+pub fn is_running() -> bool {
+    unsafe { sys::furi_kernel_is_running() }
+}
+
+/// Kernel lock state.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, uDebug)]
+pub enum LockState {
+    /// normal scheduling
+    Unlocked = 0,
+    /// scheduling paused
+    Locked = 1,
+}
+
+impl From<i32> for LockState {
+    fn from(value: i32) -> Self {
+        match value {
+            0 => LockState::Unlocked,
+            // Assume any non-zero value is `Locked``.
+            // This can be modified if new lock states are added in the future.
+            _ => LockState::Locked,
+        }
+    }
+}
+
+impl From<LockState> for i32 {
+    fn from(value: LockState) -> Self {
+        value as i32
+    }
+}
+
+/// Lock kernel, pause process scheduling.
+///
+/// <div class="warning">This should never be called in an interrupt request context.</div>
+///
+/// Returns previous lock state.
+pub fn lock() -> furi::Result<LockState> {
+    let status = sys::furi::Status::from(unsafe { sys::furi_kernel_lock() });
+
+    status.err_or_else(|status| status.0.into())
+}
+
+/// Unlock kernel, resume process scheduling.
+///
+/// <div class="warning">This should never be called in an interrupt request context.</div>
+///
+/// Returns previous lock state.
+pub fn unlock() -> furi::Result<LockState> {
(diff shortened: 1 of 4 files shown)


crates/flipperzero/src/furi/kernel.rs
```diff
@@ -0,0 +1,89 @@
...
+use crate::furi;
+
+/// Check if CPU is in IRQ; or kernel running and IRQ is masked.
+pub fn is_irq_or_masked() -> bool {
+    unsafe { sys::furi_kernel_is_irq_or_masked() }
+}
+
+/// Check if kernel is running.
+pub fn is_running() -> bool {
+    unsafe { sys::furi_kernel_is_irq_or_masked() }
```
JarvisCraft, line 15: Looks like the wrong function is called

crates/flipperzero/src/furi/kernel.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// scheduling paused
+    Locked = 1,
+}
+
+impl From<i32> for LockState {
+    fn from(value: i32) -> Self {
+        match value {
+            0 => LockState::Unlocked,
+            1 => LockState::Locked,
+            x => panic!("invalid lock state value: {x}"),
```
JarvisCraft, line 32: Panicking `From` is quite bad, since it can be called by any user. `TryFrom` with on-demand `expect` should be better.

JarvisCraft (changes requested):

crates/flipperzero/src/furi/kernel.rs
dcoles: Good catch! Copy-paste mistake.

Fixed.

crates/flipperzero/src/furi/kernel.rs
dcoles: Looking at the implementation, I believe the intent is for the `i32` to act as either an error `Status` or a boolean lock state (0: unlocked, 1: locked). Effectively, it ends up like a cast to `bool` (e.g. `if (lock_state) { // LOCKED } else { // UNLOCK }`. Hard to be sure since no code in `flipper-firmware` actually uses the results of these functions...

I've modified the behaviour to treat `0` as Unlocked and non-`0` as Locked, so it won't panic on invalid values. This would also allow more specific Locked states to be added in the future.

JarvisCraft (approved):

crates/flipperzero/src/furi/kernel.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Returns previous lock state.
+pub fn unlock() -> furi::Result<LockState> {
+    let status = sys::furi::Status::from(unsafe { sys::furi_kernel_unlock() });
+
+    status.err_or_else(|status| status.0.into())
+}
+
+/// Restore kernel lock state.
+///
+/// **warning** This should never be called in an interrupt request context.
```
JarvisCraft, line 68:
```suggestion
/// <div class="warning">This should never be called in an interrupt request context.</div>
```

Just remembered this feature: rust-lang/rust#106561

Same above

crates/flipperzero/src/furi/kernel.rs
```diff
@@ -0,0 +1,90 @@
...
+///
+/// Returns previous lock state.
+pub fn restore_lock(state: LockState) -> furi::Result<LockState> {
+    let status = sys::furi::Status::from(unsafe { sys::furi_kernel_restore_lock(state.into()) });
+
+    status.err_or_else(|status| status.0.into())
```
JarvisCraft, line 74: Just looking at the repetetive

```rust
let status = sys::furi::Status::from(...);
status.err_or_else(|status| status.0.into())
```

I start to feel that it's worth adding a convenience method to `Status` for this conversion such as this can be replaced with a one-liner

```rust
sys::furi::Status::into_result(...)
```

JarvisCraft (approved):
Adding a few non-critical suggestings

crates/flipperzero/src/furi/kernel.rs
dcoles: I agree, though I need to think about how to do it in a clean way.

The challenge with `i32` status is that it can be interpreted in many ways (there's not a single mapping).

I might be overthinking it and should just provide an `into_result(self) -> Result<i32, Status>` and let the user do the mapping as desired. `ENEEDMORECAFFEINE`. 🍵

dcoles: Since this might involving some more refactoring will address in a later PR.

PR #182, opened 2024-10-21
Provide safe wrappers for thread notification API

dcoles:
Furi provides access to [FreeRTOS Task Notification](https://freertos.org/Documentation/02-Kernel/02-Kernel-features/03-Direct-to-task-notifications/01-Task-notifications) feature which can be used for syncronization between threads.

Task notifications act as `u32` bitmask, so can be used to signal multiple potential interesting states on the target thread.


dcoles:
Accidentally merged. Rolled back `main` and created #190.

PR #186, opened 2024-10-31
Add `sys::furi::Error` type for kernel errors

dcoles:
This is an updated attempt at improving the ergonomics around working with `FuriStatus`. Rather than trying to use `sys::furi::Status` as a pseudo-`Error`/`Result` type, use a dedicated `Error` enum.

In a future PR I'm going to change how the bindings for `enum FuriStatus` to a newtype rather than integer constant. This will remove the need for a seperate `sys::furi::Status` type and the associated hoops required to turn that into a `Result` type. Being an alias for `i32` limits what traits or functions can be implemented on the type.

### DIFF
--- a/crates/flipperzero/src/furi/message_queue.rs
+++ b/crates/flipperzero/src/furi/message_queue.rs
@@ -40,7 +40,9 @@ impl<M: Sized> MessageQueue<M> {
             .into()
         };
 
-        status.err_or(())
+        let _ = status.into_result()?;
+
+        Ok(())
     }
 
     // Attempts to read a message from the front of the queue within timeout ticks.
@@ -55,11 +57,9 @@ impl<M: Sized> MessageQueue<M> {
             .into()
         };
 
-        if status.is_ok() {
-            Ok(unsafe { out.assume_init() })
-        } else {
-            Err(status)
-        }
+        let _ = status.into_result()?;
+
+        Ok(unsafe { out.assume_init() })
     }
 
     /// Returns the capacity of the queue.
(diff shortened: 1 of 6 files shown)


dcoles:
Follow-up to #183.

crates/flipperzero/src/furi/message_queue.rs
```diff
@@ -55,11 +55,7 @@ impl<M: Sized> MessageQueue<M> {
             .into()
         };
 
-        if status.is_ok() {
-            Ok(unsafe { out.assume_init() })
-        } else {
-            Err(status)
-        }
+        status.into_result().map(|_| unsafe { out.assume_init() })
```
JarvisCraft, line 58: IMO, this is the case where `match` looks better (given `_` binding and `unsafe`), but this may just be a style preference

JarvisCraft (approved):
Good job!

Adding a small non-blocking note, just in case

crates/flipperzero/src/furi/message_queue.rs
dcoles: So like the following?

```rust
match status.into_result() {
    Err(err) => Err(err),
    Ok(_) => Ok(out.assume_init()),
}
```

I guess using an `if let` expression would also work here:

```rust
if let Err(err) = status.into_result() {
    Err(err)
} else {
    unsafe { out.assume_init() }
}
```

or even:

```rust
let _ = status.into_result()?;

unsafe { out.assume_init() }
```

dcoles: I think that last one is the most explicit that we're not using the actual status code for anything.

I updated the code to match that style.

PR #194, opened 2025-01-04
sys: Fix `halt!` and `crash!` message handling

dcoles:
Due to the way we called `__furi_crash_implementation` the value of the `r12` register would often get clobbered, corrupting the message.

This is fixed by directly jumping to the crash handler from inline assembly.

Only messages stored on flash memory can be shown at reboot, so strings stored in RAM (including string constants in external apps) will be replaced by a generic "check serial logs" message.

This also allows calling `halt!` and `crash!` in their no-argument form.

### DIFF
--- a/crates/sys/src/lib.rs
+++ b/crates/sys/src/lib.rs
@@ -34,8 +34,8 @@ mod bindings;
 
 /// Crash the system.
 ///
-/// The only argument is a message with which the system should crash
-/// which should contain no NULs. The following will not compile:
+/// May be provided with an optional message which should not contain NULs.
+/// The following will not compile:
 ///
 /// ```compile_fail
 /// flipperzero_sys::crash!("Has a \0 NUL");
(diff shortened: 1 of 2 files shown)


crates/sys/src/lib.rs
```diff
@@ -48,21 +48,79 @@ mod bindings;
...
             };
+        )?
+
+        unsafe {
+            // Crash message is passed via r12
+            ::core::arch::asm!(
+                "ldr pc,=__furi_crash_implementation",
+                in("r12") message,
+                options(nomem, nostack))
+            ;
```
JarvisCraft, line 75:
```suggestion
                options(nomem, nostack),
            );
```

Feels  like a more rustfmt'ish  formatting

crates/sys/src/lib.rs
```diff
@@ -48,21 +48,79 @@ mod bindings;
...
-            match ::core::ffi::CStr::from_bytes_with_nul(
-                ::core::concat!($msg, "\0").as_bytes(),
-            ) {
-                Ok(cstr) => cstr.as_ptr(),
-                Err(error) => panic!("message contains NULs"),
+    ($($msg:expr)? $(,)?) => {{
+        // Optional message
+        let mut message: *const i8 = ::core::ptr::null();
+        $(
+            message = match ::core::ffi::CStr::from_bytes_with_nul(::core::concat!($msg, "\0").as_bytes()) {
```
JarvisCraft, line 63: This is out of  this PR's scope, but  it may be worth switching to `CStr` for `$msg`

crates/sys/src/lib.rs
```diff
@@ -48,21 +48,79 @@ mod bindings;
...
-    ($msg:expr $(,)?) => {{
-        const MESSAGE: *const ::core::primitive::i8 =
-            match ::core::ffi::CStr::from_bytes_with_nul(
-                ::core::concat!($msg, "\0").as_bytes(),
-            ) {
-                Ok(cstr) => cstr.as_ptr(),
-                Err(error) => panic!("message contains NULs"),
+    ($($msg:expr)? $(,)?) => {{
+        // Optional message
+        let mut message: *const i8 = ::core::ptr::null();
```
JarvisCraft, line 61: I am  not a  big fan  of this maybe-redundant initialization, so I would  prefer this to (still)  be  a  `($msg:expr $(,)?)` branch with an empty branch `()` delegating to it
Also note that this  would fix  the  (currently added) problem of `crash(,)` being a valid expression.

crates/sys/src/lib.rs
```diff
@@ -48,21 +48,79 @@ mod bindings;
...
+/// ```
+///
+/// Halt the system with default *"System halt requested."* message:
+///
+/// ```
+/// flipperzero_sys::crash!();
+/// ```
+#[macro_export]
+macro_rules! halt {
+    ($($msg:expr)? $(,)?) => {{
```
JarvisCraft, line 106: All the concerns mentioned above are valid here too

JarvisCraft (changes requested):

crates/sys/src/lib.rs
dcoles: That's odd. I agree. I wonder if rustfmt did something weird on me or if it was just me. 😅

crates/sys/src/lib.rs
dcoles: I was wondering about that too. Certainly much less painful now that there's `CStr` literals.

(Funny enough, it's really hard to see these messages without using the hardware UART. Usually the USB connection drops before the message gets written to the USB TTY).

crates/sys/src/lib.rs
dcoles: Sounds good. I agree that's a better approach.

dcoles: Since the crash handler specifically requires a pointer, I added a `__crash/halt_implementation!` macro that takes the pointer from the public `crash/halt!` macros rather than resorting to pattern-matching trickery.

crates/sys/src/lib.rs
```diff
@@ -48,26 +48,103 @@ mod bindings;
...
+#[macro_export]
+macro_rules! halt {
+    () => {
+        $crate::__halt_implementation!(::core::ptr::null());
+    };
+    ($msg:expr $(,)?) => {{
+        // Optional message
+        let message =
+            ::core::ffi::CStr::from_bytes_with_nul(::core::concat!($msg, "\0").as_bytes())
+                .unwrap_or(c"nul in halt message");
```
JarvisCraft, line 122: Shouldn't it be wrapped in `const {...}` to ensure CT evaluation?

JarvisCraft (changes requested):

crates/sys/src/lib.rs
dcoles: I did not know about that. Cool!

dcoles: That also caught that `unwrap_or` isn't `const` and so I need to use `match`.

JarvisCraft (approved):

dcoles:
@JarvisCraft Thank you for the quick turnaround on the PRs! Much appreciated! 🎉

PR #204, opened 2025-01-24
Fix display of panic message

dcoles:
When connected via USB serial, there is often not enough time for the output to be fully recieved before the USB device resets.

Add a short 500 ms delay before we call `furi_crash`.

This also fixes incorrect `printf` formatting strings with explicit length (the length is provided as the 'precision' value after the decimal rather than the 'width' value before the decimal).


dcoles:
I'm going to merge this for the release, since it fixes a pretty nasty bug.

Issue #205, opened 2025-01-25
Update to Rust Edition 2024


loftyinclination:
This is landing soon, and it'd maybe be nice to update. Not sure what effect updating would have on the msrv, but the 2024 edition does come with a few nice features for embedded code and FFI (in particular, declaring [functions in `unsafe extern` blocks as `safe`](https://dev-doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html)).

I've tried doing this update locally and found a few minor issues that were easily fixed; 
* [`gen` is now a reserved keyword](https://dev-doc.rust-lang.org/edition-guide/rust-2024/gen-keyword.html), and so a few of the bindings need escaping.
* [linker attributes are now unsafe and need to be marked as such](https://dev-doc.rust-lang.org/edition-guide/rust-2024/unsafe-attributes.html). Think this is probably safe to do?
* extern blocks are now required to be marked as unsafe, and a number of the blocks in the v0.13 bindings weren't
* [the `unsafe_op_in_unsafe_block` lint is now a warning by default](https://dev-doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html); this is pretty easily solved, but may just warrant the lint being disabled.

Would be interested to hear your thoughts on updating to this new edition. I've got a local branch that I could raise a draft PR for if that would be helpful.

loftyinclination:
Not sure if this is a local issue, but after changing edition to 2024, a number of the string conversion functions (panic, logging, and some furi string methods) fail to compile due to expecting a u8 pointer but instead receiving i8 pointers.

`error[E0308]: mismatched types
    --> flipperzero/examples/i2c-ds3231.rs:79:9
     |
79   |         error!("DS3231 is not connected and ready");
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |         |
     |         expected `*const u8`, found `*const i8`
     |         arguments to this function are incorrect
     |
     = note: expected raw pointer `*const u8`
                found raw pointer `*const i8`
note: function defined here
    --> /home/lofty/Documents/Creative/Programming/Rust/flipper-zero-api/crates/sys/src/bindings.rs:1968:12
     |
1968 |     pub fn furi_log_print_format(
     |            ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #209, opened 2025-01-28
Address some 2024 compatibility issues

loftyinclination:
As suggested in https://github.com/flipperzero-rs/flipperzero/issues/205, this enables the `rust_2024_compatibility` lints, so that we can see what is necessary to fix before updating.

In the issue, two blockers in `bindgen` were mentioned; the first of these, `unsafe extern` blocks, has landed, and this patch additionally updates to latest `bindgen` and enables this behaviour.


loftyinclination:
A large chunk of the warnings introduced by this are due to the `edition-2024-expr-fragment-specifier` lint, and I think it's likely that those cases are probably fine, and that we could suppress that lint.

dcoles:
Thanks for the PR! This looks great. I'll trigger regeneration of the bindings once this is merged which should substantially reduce the number of warnings in `sys`.

dcoles (approved):

dcoles:
> Thanks for the PR! This looks great. I'll trigger regeneration of the bindings once this is merged which should substantially reduce the number of warnings in sys.

Whoops. I see you've already regenerated the bindings. I'll still trigger the action just to make sure it generated cleanly.

loftyinclination:
Ah, apologies for regenerating the bindings.

That's unfortunate about the docker dependency. Hopefully there's something we can do.

I've got another local change to address some of the warnings, I'll push it up later today.

PR #215, opened 2025-02-01
Additional Records

dcoles:
This PR adds `Clone` support to `UnsafeRecord` so that it can be used for reference-counting patterns and adds `Clone` for the existing Record handles.

Additionally it adds handles for `Power` and `Storage` services.


dcoles:
I'm going to merge this since there's been no comments so far, but still open for post merge review.

PR #216, opened 2025-02-02
Add serial support.

dcoles:
This provides access to the serial interface using an async worker to safely transfer received bits from the interrupt handler to the application.

Add a simple serial echo example and wrapper for `FuriEventFlag` that can be used to wait for an event without polling.


dcoles:
I'm going to merge this since there's been no comments so far, but still open for post merge review.

PR #223, opened 2025-02-28
Update to 2024 Edition

dcoles:
This has been tested, but can't be merged until there is a new `bindgen` release with the changes from https://github.com/rust-lang/rust-bindgen/pull/3100 that rename use of the now-reserved `gen` keyword.


pauliesnug:
How many years should we wait for a new `bindgen` release before we just go with a random git commit hash /hj

dcoles:
> How many years should we wait for a new `bindgen` release before we just go with a random git commit hash /hj

Unfortunately it's not possible to publish a crate to crates.io with a git dependency. We could certainly do so on a branch for testing, it just couldn't be part of an official release.

PR #230, opened 2025-05-21
Rebase Bluetooth APIs PR (#139)

pauliesnug:
stale since 2024, fixed code with new API changes. should be mergable! :D

### DIFF
--- a/crates/flipperzero/src/bluetooth/beacon.rs
+++ b/crates/flipperzero/src/bluetooth/beacon.rs
@@ -1,0 +1,661 @@
+//! Bluetooth beacon.
+
+use flipperzero_sys as sys;
+use rand_core::RngCore;
+use ufmt::derive::uDebug;
+use uuid::Uuid;
+
+use crate::furi::rng::HwRng;
+
+/// The maximum length of the data broadcast by the beacon.
+pub const BEACON_MAX_DATA_SIZE: usize = 31;
+
+/// The Bluetooth beacon that can be broadcast from the Flipper Zero.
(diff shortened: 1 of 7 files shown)


crates/flipperzero/src/bluetooth/beacon.rs
```diff
@@ -0,0 +1,661 @@
...
+    pub fn eddystone_url(tx_power: u8, url_scheme: EddystoneUrlScheme, url: &str) -> Self {
+        let mut service_data = alloc::vec![
+            0xAA, // Eddystone ID
+            0xFE,
+            0x10, // Frame type
+            tx_power,
+            url_scheme.to_number(),
+        ];
+
+        let mut url_bytes = url.as_bytes();
```
dcoles, line 367: This matching should be done on a percent-encoded URL string to ensure that the string doesn't include any reserved bytes (e.g. `' '`). The [`url` create](https://crates.io/crates/url) might be suitable here.

dcoles (changes requested):

crates/flipperzero/src/bluetooth/beacon.rs
pauliesnug: Do we even need the advertisement packets? It's not a part of the C API anyways...

pauliesnug:
@dcoles Fixed almost everything, but I'm wondering if we need the AdPacket etc., as the the Eddystone API is seemingly deprecated and I've never heard of a usecase for those other beacon types...

PR #236, opened 2025-09-04
Fix lifetime of `ExclusiveCanvas` canvas references

dcoles:
To get mutable access to a `Canvas` reference you must first call `gui_direct_draw_acquire` then release it with `gui_direct_draw_release` when done.

In our Rust wrapper, this is enforced by returning an `ExclusiveCanvas` handle which represents the claim and can be use for borrowing a `Canvas`.

As such, the following should not compile (temporary value dropped while borrowed):

```rust
let canvas: &mut Canvas = gui.direct_draw_acquire().canvas_mut();
````

However, before this fix it did, due to the borrow happening on `Gui` and not `ExclusiveCanvas`. This would make it impossible to actually write to the Canvas, which was part of the issue in #235.

`ExclusiveCanvas` uses `PhantomData` to hold a "reference" to `Gui`, so by holding a reference to `ExclusiveCanvas` we transitively also hold a reference to `Gui`, thus don't need to explicitly include `Gui`'s lifetime in returned references.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 - Switched to `nightly-2025-08-31` compiler
+- Fix lifetime of `ExclusiveCanvas::canvas` and `canvas_mut`
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


JarvisCraft (approved):

JarvisCraft:
Could you add a `compile_fail` doctest for the fix?

PR #250, opened 2026-01-04
Get Miri functioning by reimplementing all of `sys` in Rust

loftyinclination:
This is the start of some work on attempting to [get MIRI working](https://github.com/flipperzero-rs/flipperzero-rs/issues/100). Since MIRI doesn't support FFI (outside of simple methods that don't allocate and that don't deal with pointers), this PR instead follows a similar path to the one put forward by Loom; namely, to define alternate versions of all of the problematic methods and types, and swap to use these types when running under the MIRI configuration.

Current status of work;
 * the alternate types and methods have all been defined, but are currently all method implementations have been left as TODOs.
 * The configuration for the project has been changed such that both `cargo check` and `cargo miri run --example {gui,view_dispatcher}` complete without any errors.
 * A few alternate implementations have been added for glue logic (such as for the global allocator and the panic handler), based on the examples laid out in the MIRI project's test directories.

Necessary remaining work;
 * reproducing the behaviour of the C implementations in the `sys/miri_bindings` types.
 * allowing for more of the examples to be run under this scope
 * adding some more examples to test some specifics.

Got started on this because I have been rebasing the [gui-bindings](https://github.com/flipperzero-rs/flipperzero-rs/pull/29) PR onto main, and getting that back into working order. I might end up rebasing these changes on top of that local work (which I may also push up, if there's an interest), as that's really where I'd like to benefit from MIRI's analysis.


crates/flipperzero/Cargo.toml
```diff
@@ -78,6 +79,7 @@ default = ["embedded-graphics"]
 ## extern crate flipperzero_alloc;
 ## ```
 alloc = []
+not-miri = []
```
dcoles: It seems a bit odd to support a `miri` and `not-miri` feature at the same time.

I know we could use `#[cfg_attr(miri, ignore)]` on all unsupported test cases, but that's also pretty inconvenient.

dcoles (commented):
I'd be pretty excited to see this working since working with C FFI is one of the biggest challenges for this project.

My biggest concern would be the long-term maintenance of all those stubs. The Flipper Zero SDK exposes over 2000 public functions and it's not unheard of for them to break parts of the API across major firmware updates.

We might want to be a bit targeted here, such as providing coverage for the GUI APIs rather than trying to tackle all of `sys` in one go. I don't see that being maintainable without some support from bindgen (especially if we were to switch to generating `sys` on the fly, see #173).

loftyinclination:
> We might want to be a bit targeted here, such as providing coverage for the GUI APIs rather than trying to tackle all of `sys` in one go. 

I agree -- I've mostly tried to be limited in what I'm reimplementing here (the title of this PR is probably a little misleading in that regard). 

> The Flipper Zero SDK exposes over 2000 public functions and it's not unheard of for them to break parts of the API across major firmware updates.

The current changes here only define symbols in `sys` that are actually used, and so I don't imagine that this would introduce much extra surface area for breaking changes than would otherwise be incurred by utilising the `extern C` types?

loftyinclination:
This kinda almost works! Couple of problems to solve, namely 6a1fa99 and e29e99c, but I figured that it was worth pushing this up for others to look at.

loftyinclination:
I've managed to bodge together a command for running only those examples that should be run under MIRI; 
```
cargo metadata --no-deps --format-version=1 | jq '.packages[] | select(.name == "flipperzero") | .targets[] | select((.kind | contains(["example"])) and (.doc | not) and ((."required-features" | not) or (."required-features" | contains(["not-miri"]) | not))).name' | xargs -i sh -c 'cargo miri run --features alloc --target thumbv7em-none-eabihf --example {}'
```

loftyinclination:
I consider this now ready to go in -- it's been condensed a lot, so that all of the logical implementation of the MIRI logic is in basically one commit, with the individual precursor work to get MIRI able to run at all (turning off unimplemented modules, disabling examples that we're not intending to run) in their own standalone commits, and then finally work to get the examples to run under MIRI on CI  in the final few commits. 

CI checks the output of `cargo metadata` in order to find which examples to run, to save the trouble of maintaining a list of examples that we want to be able to test in the future.

Now that I'm happy enough with this, I'll get started on splitting out the GUI changes, if that would still be welcome.

crates/flipperzero/examples/gui.rs
```diff
@@ -43,16 +43,74 @@ fn main(_args: Option<&CStr>) -> i32 {
...
         {
-            let gui = UnsafeRecord::open(c"gui");
+            let gui: UnsafeRecord<Gui> = UnsafeRecord::open(c"gui");
+            #[cfg(miri)]
+            debug_assert_eq!(
+                {
+                    extern crate alloc;
+                    use alloc::sync::Arc;
+
+                    let gui: Arc<Gui> = Arc::from_raw(gui.as_ptr());
```
dcoles: The contract for [`Arc::from_raw`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.from_raw) specifically says:
> The raw pointer must have been previously returned by a call to [Arc<U>::into_raw](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_raw) or [Arc<U>::into_raw_with_allocator](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_raw_with_allocator).

However this pointer is allocated by the C runtime, so would normally be undefined behaviour.

Is this a typical pattern for Miri to model a foreign ref-counted value? If so, a comment would be helpful.

crates/flipperzero/examples/gui.rs
```diff
@@ -43,16 +43,74 @@ fn main(_args: Option<&CStr>) -> i32 {
...
     }
 
     0
 }
+
+#[cfg(miri)]
+#[unsafe(no_mangle)]
+fn miri_start(_argc: isize, _argv: *const *const u8) -> isize {
+    // TODO: Is there any benefit to Miri in hooking up the binary arguments to
+    // the test runner?
```
dcoles: Since these programs ignore any command line arguments, it's fine to do the same for the Miri arguments.

crates/flipperzero/Cargo.toml
dcoles: I think a better solution would be to add `#![cfg(not(miri))]` to the top of each of the tests, rather than introduce a `not-miri` feature.

crates/sys/src/furi/alloc.rs
```diff
@@ -26,6 +26,15 @@ impl<T> FuriBox<T> {
...
 
+    #[cfg(miri)]
+    /// Allocates and initializes a correctly aligned value on the system heap.
+    pub fn new(value: T) -> Self {
+        unsafe {
+            // SAFETY: Pointer is non-null, aligned and represents a valid `T`
+            FuriBox::from_raw(value)
+        }
+    }
+
```
dcoles, line 37: It looks like this definition would conflict with the existing `FuriBox::new` if the feature was ever enabled.

dcoles (changes requested):
This all looks pretty good to me. I'd be happy to merge once the comments are addressed.

crates/flipperzero/Cargo.toml
loftyinclination: That would prevent us from differentiating at the manifest level between examples that support running under MIRI and those that don't, which I don't especially like for CI?

I do agree that this is an unfortunate addition though.

(Additionally, calling "cargo miri run" on a test with `#![cfg(not(miri))]` at the top (after `#![no_std] #![no_main]`) complains about a lack of panic handler.)

crates/flipperzero/examples/gui.rs
loftyinclination: > Is this a typical pattern for Miri to model a foreign ref-counted value? 

I'm not sure about a typical pattern, but it's what I've gone with for the mocking of the C runtime structs in this PR. I definitely feel that it's important to be able to test the threading as part of this work, since that was one of the areas where I was getting tripped up.

> A comment would be helpful.

I've now extracted these checks into a macro that does this check, and added a comment to explain the safety and reasoning.

dcoles: That seems reasonable. My impression is that it just has to look like it was implemented in Rust from Miri's perspective, not actually be functional.

PR #261, opened 2026-03-19
Rebase GUI APIs, bring up to date, and test with MIRI

loftyinclination:
This is a continuation of the work in https://github.com/flipperzero-rs/flipperzero-rs/pull/250 and in https://github.com/flipperzero-rs/flipperzero-rs/pull/29, rebasing the latter onto the former and continuing the work of creating some useful GUI bindings.

Differences from the GUI PR;
* Callbacks have been changed to be more idomatic in their return type where possible; enums are returned and the FFI glue converts that into the relevant type for the Flipper codebase.
* View bindings have now been properly added
* The View Dispatcher struct has been simplfied somewhat, now that the event queue is always enabled.
* Safe wrappers around the Submenu and Variable Item List structs have been added
* Examples have been added (mostly to enable MIRI testing to take place).

I'd consider this a good starting point, though there's likely still a lot of work left to go on making this more ergonomic, and documenting all of the safety guarantees.

Known issues;
* The View Dispatcher example does not run under MIRI, because of an ABBA lock issue.


crates/flipperzero/src/gui/submenu.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[cfg(feature = "alloc")]
+impl<'gui, VDC: ViewDispatcherCallbacks> SubmenuBoundToViewDispatcher<'gui, VDC> {
+    /// Adds a new item to the submenu that, when interacted with, will switch the
+    /// `ViewDispatcher`'s current `View` to the one provided to this method.
+    pub fn add_nav_item<'label, VC: ViewCallbacks>(
+        &mut self,
+        label: &'label CStr,
+        // TODO: allow this to take a view id?
+        view: &ViewDispatcherView<'gui, VC, VDC>,
```
loftyinclination, line 190: this should take `dyn AsRef<ViewDispatcherView>` so that it can take eg `VariableItemListBoundToDispatcher`

crates/flipperzero/src/gui/variable_item_list.rs
```diff
@@ -0,0 +1,474 @@
...
+use alloc::{boxed::Box, vec::Vec};
+use core::mem::MaybeUninit;
+use core::ops::{Deref, DerefMut};
+use core::{
+    ffi::c_void,
+    ptr::{self, NonNull},
+};
+use flipperzero_sys as sys;
+
+pub struct VariableItemList<'a, T> {
```
loftyinclination, line 18: Ideally, this type would be `Send` (especially useful since other threads might want to add new items to the list).

It's not automatically because it contains a pointer to the `flipperzero_sys::VariableItemList`, and because it holds a `Arc<Mutex<CallbackContextInner>>` (the Arc is required so that the context may be read from within callbacks, which are invoked by input events, which happen on the ViewDispatcher::run thread).

loftyinclination:
@dcoles was wondering if you had any time to spend on discussing a direction to potentially landing this?

dcoles:
> @dcoles was wondering if you had any time to spend on discussing a direction to potentially landing this?

I'd certainly like to do so. The biggest challenge is that 9k+ lines of code and 100+ commits makes it pretty daunting to review in one go.

Usually it's best to try to break it down into smaller pieces that can be reviewed and committed independently. That helps avoid the whole PR getting caught up because of discussion about a specific area of code. That's part of the reason the original GUI PR never landed (though we did pull some pieces in piecemeal).

The Miri changes look like the biggest, so it would be nice to start with the minimum support required, such as running it under GitHub actions (possibly this is what #250 was intended to be). At present the "Build and test" step is failing which I'd also consider a blocker.

If there's any updates required to `sys/src/bindings.rs` we usually like to do those via [update_bindings workflow](https://github.com/flipperzero-rs/flipperzero-rs/actions/workflows/update_bindings.yml) so that the changes are trackable rather than mixed in with other changes.

That said, I'd be happy to take a look over the current PR and see if I can spot anything of major concern.

loftyinclination:
> I'd certainly like to do so. The biggest challenge is that 9k+ lines of code and 100+ commits makes it pretty daunting to review in one go.

Very understandably! I'm happy to work on getting this split up into some smaller PRs (apologies if my message made it seem like I was asking you to look at the whole thing in one go!).

From my perspective, the logical grouping would probably be

1. Miri (with examples that support Miri running on CI)
2. the input event module (as precursor work that stands mostly on its own, and because it's necessary to react to any event)
3. the GUI module and maybe the ViewPort module (the GUI record is necessary for hooking anything into the GUI, and the ViewPort is the lowest primitive that you might want to use, if you've got a really simple application to write).
4. the view dispatcher and view modules (as the primitives that you'd usually interact with)
5. the submenu or variable item list modules (as a start on adding safe wrappers for the flipper intrinsic views)

and any updates to the bindings could be done at any point in that.

I'll put any further questions about the MIRI work in #250

loftyinclination:
I consider this now ready to go in -- it's been condensed a lot, so that all of the logical implementation of the MIRI logic is in basically one commit, with the individual precursor work to get MIRI able to run at all (turning off unimplemented modules, disabling examples that we're not intending to run) in their own standalone commits, and then finally work to get the examples to run under MIRI on CI  in the final few commits. 

CI checks the output of `cargo metadata` in order to find which examples to run, to save the trouble of maintaining a list of examples that we want to be able to test in the future.

Now that I'm happy enough with this, I'll get started on splitting out the GUI changes, if that would still be welcome.

loftyinclination:
discount _all_ of that last message, I'm looking at the wrong PR.

PR #272, opened 2026-07-24
FuriString: Avoid use of `furi_string_reset`

### LINKED ISSUE #271: FuriString::clear frees the backing structure, invalidating any pointers
The FFI implementation for `furi_string_reset()` internally calls `string_clear()`, which clears the sys::FuriString's backing pointer. 

As such, the following is a use-after-free;

```rust
# use flipperzero::furi::string::FuriString;

let a: FuriString = c"String".into();
let b = a.as_c_ptr();
a.reset();
unsafe { b.read() };
```

dcoles:
As of flipperdevices/flipperzero-firmware@a46038acbf576a3102c6d89c0f142c537e3cae1f `furi_string_reset` now calls `string_clear` + `string_init` rather than `string_reset`.

This results in the string capacity being reallocated rather than just truncated to zero length, resulting in existing pointers being invalidated.

Fixes: #271

### DIFF
--- a/crates/flipperzero/src/furi/string.rs
+++ b/crates/flipperzero/src/furi/string.rs
@@ -89,6 +89,9 @@ impl FuriString {
 
     /// Extracts a pointer to a raw zero-terminated string
     /// containing the entire string slice.
+    ///
+    /// The pointer remains valid until the next call that
+    /// modifies the string or the string is dropped.
     #[inline]
     #[must_use]
     pub fn as_c_ptr(&self) -> *const c_char {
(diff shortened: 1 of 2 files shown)


crates/flipperzero/src/furi/string.rs
loftyinclination, line 90: It's also probably worth documenting that this pointer is only valid under some writes to the string

dcoles: That's a good suggestion. Thank you.

PR #274, opened 2026-08-17
Check all targets in clippy now that issue 117 is resolved

loftyinclination:
This has been resolved since v1.3.0 of auguwu/clippy-action, which dependabot
helpfully updated us to a few years ago.


loftyinclination:
Hmm looks like this would require enabling the "check: write" permission

JarvisCraft:
> Hmm looks like this would require enabling the "check: write" permission

@loftyinclination, feel free to add it to the corresponding `YAML`-files.

loftyinclination:
I've added the `permissions` entry to the job, but it's still not working 
```Warning: clippy-action doesn't have permissions to create Check Runs, disabling!```

Additionally, looks like using "args: --all-targets" results in a clippy command like `cargo clippy -- --all-targets`, which clippy doesn't like -- it needs --all-targets to be specified before the --

loftyinclination:
Changed `args` to `check-args` as per [issue 495](https://github.com/auguwu/clippy-action/issues/495)

JarvisCraft:
Thanks! As for the reported problems, I may have tie ob this week to try and resolve them (this also includes updating some dependencies)

PR #275, opened 2026-08-17
remove unused digest dependency

loftyinclination:
This has been unused since the MD5 and SHA256 modules were removed in 48f0f379dd3f27747efa63e5b99641904e3d922b.


crates/flipperzero/Cargo.toml
JarvisCraft, line 1: Looks like `Cargo.lock` has not been regenerated

JarvisCraft (changes requested):

crates/flipperzero/Cargo.toml
loftyinclination: crates/flipperzero doesn't have a Cargo.lock file, and while there might be one at crates/Cargo.lock, it's not checked in

JarvisCraft: My bad, looked in the wrong `Cargo.toml`

